PR #30, opened 2022-10-11
Fix SELECT count(*)

akuzm:
This query only requires counting rows, so it is optimized into selecting a rowid column from the underlying table. In case of the Postgres table, this gets deparsed into selecting the first column. If it happens to be a text column, we then try to insert it into the rowid output vector which is int, and fail with an assertion because of type mismatch.

Instead of this, deparse this case into SELECT NULL FROM postgres_table. This is enough for counting the rows.


Mytherin:

PR #31, opened 2022-10-12
Interval support

hannes:


Makefile
```diff
@@ -40,4 +40,3 @@ format:
 	clang-format --sort-includes=0 -style=file -i postgres_scanner.cpp
 	clang-format --sort-includes=0 -style=file -i concurrency_test.cpp
```
Mytherin, line 41: Perhaps we should include the same `.clang-format` file as used in the main repo?

hannes: It probably should you're right

PR #43, opened 2022-10-27
support uuid type

### LINKED ISSUE #25: unsupported type uuid
duckdb version: `v0.5.1 7c111322d`

To reproduce:
```sql
-- in postgres
CREATE TABLE example (
  id uuid
);
```

```
-- in duckdb
D install postgres_scanner;
D load postgres_scanner;
D select * from postgres_scan('dbname=db', 'public', 'example') limit 10;
Error: Invalid Error: IO Error: Unsupported Postgres type uuid
```

I have a [patch](https://github.com/lepfhty/postgres_scanner/commit/b43c027893e00f9a47ab81ae665e5da0fb89acfe) that seems to be working. I updated a test and got `make test` to pass locally (with some changes to the build/test scripts that are not included in the patch). 

Love what you guys are doing. I can open a pull request if you'd like.

lepfhty:
Uses `hugeint_t` data type to represent UUID from postgres.
Adds test cases for "nil uuid", lowercase, uppercase, and `NULL`.
Fixes #25

### DIFF
--- a/postgres_scanner.cpp
+++ b/postgres_scanner.cpp
@@ -196,6 +196,8 @@ static LogicalType DuckDBType(PostgresColumnInfo &info, PGconn *conn) {
 		return LogicalType::TIMESTAMP_TZ;
 	} else if (pgtypename == "interval") {
 		return LogicalType::INTERVAL;
+	} else if (pgtypename == "uuid") {
+		return LogicalType::UUID;
 	} else {
 		throw IOException("Unsupported Postgres type %s", pgtypename);
 	}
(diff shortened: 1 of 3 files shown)


postgres_scanner.cpp
```diff
@@ -715,6 +717,22 @@ static void ProcessValue(data_ptr_t value_ptr, idx_t value_len, const PostgresBi
...
+
+	case LogicalTypeId::UUID: {
+		D_ASSERT(bind_data->columns[col_idx].attlen == 2 * sizeof(int64_t));
+		D_ASSERT(value_len == 2 * sizeof(int64_t));
+		D_ASSERT(bind_data->columns[col_idx].atttypmod == -1);
+
+		hugeint_t res;
+
+		auto upper = ntohll(Load<uint64_t>(value_ptr));
+		res.upper = upper ^ (int64_t(1) << 63);
```
lepfhty, line 729: I'm not sure why this is needed. I copied this from https://github.com/duckdb/duckdb/blob/master/src/common/types/uuid.cpp#L52

hannes:
thanks!

PR #46, opened 2022-11-01
add jsonb, 1D array, UUID and unbound NUMERIC support

alitrack:
uuid implementation is from [lepfhty](https://github.com/lepfhty)

### DIFF
--- a/postgres_scanner.cpp
+++ b/postgres_scanner.cpp
@@ -4,6 +4,9 @@
 #include <libpq-fe.h>
 
 #include <arpa/inet.h>
+#include <iostream>
+#include <regex>
+
 // htonll is not available on Linux it seems
 #ifndef ntohll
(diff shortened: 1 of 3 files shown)


postgres_scanner.cpp
```diff
@@ -15,6 +18,66 @@
 #include "duckdb/planner/filter/constant_filter.hpp"
 
 using namespace duckdb;
+int32_t numeric_width, numeric_scale;
+// jsonb starts with "\x01", remove it.
+std::string trim(std::string str) {
+	return regex_replace(str, std::regex("(\\x01)"), "");
```
Mause, line 24: Why are you using regex to trim a single char?

hannes:
Thanks, we merged #43 so this would need some rebasing. Could you have a go at this please?

PR #48, opened 2022-11-01
Remove LogicalType::JSON from switch cases

lnkuiper:
Removing these for my eventual PR to DuckDB where I move the JSON LogicalType into the JSON extension.


hannes:
thanks

PR #51, opened 2022-11-08
1D array and unbound NUMERIC support

alitrack:


hannes:
I've ended up re-implementing array support in a generic way. Thanks so much for the investigation work, it was very helpful getting this done quickly. For the NUMERIC support, I think it's best converting those to a DOUBLE. Working on this now.

PR #59, opened 2022-11-23
Bypass CTID filter for full range

FranckPachot:
When `pgclass.relpages=0` the filtering on CTID covers the whole range from `0` to `POSTGRES_TID_MAX`. In this case, the where clause ` between '(0,0)'::tid and '(4294967294,0)'::tid` is useless. This patch removes this filter from the where clause when in this situation.

There are two goals for this:
- Better performance when querying PostgreSQL as the useless filter takes CPU resources. It also triggers Parallel Seq Scany which may overload the server as the calls are already multithreaded.
- Allows to query YugabyteDB (Open Source Distributed SQL Postgres-compatible database) which doesn't support CTID because the storage is not in heap tables but distributed LSM-Tree

All this is detailed in: https://dev.to/franckpachot/duckdb-on-yugabytedb-31l

### DIFF
--- a/postgres_scanner.cpp
+++ b/postgres_scanner.cpp
@@ -538,32 +538,32 @@ static PostgresDecimalConfig ReadDecimalConfig(const_data_ptr_t &value_ptr) {
 };
 
 template <class T>
-static T ReadDecimal(idx_t scale, int32_t ndigits, int32_t weight, bool is_negative, const_data_ptr_t value_ptr) {
+static T ReadDecimal(PostgresDecimalConfig &config, const_data_ptr_t value_ptr) {
 	// this is wild
-	auto scale_POWER = POWERS_OF_TEN[scale];
+	auto scale_POWER = POWERS_OF_TEN[config.scale];
 
-	if (ndigits == 0) {
+	if (config.ndigits == 0) {
 		return 0;
 	}
 	T integral_part = 0, fractional_part = 0;
 
-	if (weight >= 0) {
-		D_ASSERT(weight <= ndigits);
+	if (config.weight >= 0) {
+		D_ASSERT(config.weight <= config.ndigits);
 		integral_part = LoadEndIncrement<uint16_t>(value_ptr);
-		for (auto i = 1; i <= weight; i++) {
+		for (auto i = 1; i <= config.weight; i++) {
 			integral_part *= NBASE;
-			if (i < ndigits) {
+			if (i < config.ndigits) {
 				integral_part += LoadEndIncrement<uint16_t>(value_ptr);
 			}
 		}
 		integral_part *= scale_POWER;
 	}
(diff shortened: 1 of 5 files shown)


postgres_scanner.cpp
```diff
@@ -397,13 +397,25 @@ static void PostgresInitInternal(ClientContext &context, const PostgresBindData
 		}
 		filter_string = " AND " + StringUtil::Join(filter_entries, " AND ");
```
stevefan1999-personal, line 398: Remove the " AND " prefix

postgres_scanner.cpp
```diff
@@ -397,13 +397,25 @@ static void PostgresInitInternal(ClientContext &context, const PostgresBindData
...
 	}
+        if ( task_min == 0 && task_max == POSTGRES_TID_MAX ) {
+          // no use of CTID when task_min / task_max cover the largest range. 
+	  // This includes PostgreSQL compatible database which do not provide CTID (like YugabyteDB)
+          lstate.sql = StringUtil::Format(
+            R"(
+COPY (SELECT %s FROM "%s"."%s" WHERE 1=1                                          %s) TO STDOUT (FORMAT binary);
```
stevefan1999-personal, line 405: So that you can remove "1=1" which is pretty ugly imo

postgres_scanner.cpp
```diff
@@ -397,13 +397,25 @@ static void PostgresInitInternal(ClientContext &context, const PostgresBindData
...
+)",
+
+            col_names, bind_data->schema_name, bind_data->table_name,                     filter_string);
 
-	lstate.sql = StringUtil::Format(
+	} else {
+          // use of CTID on the task_min / task_max range.		
+	  lstate.sql = StringUtil::Format(
 	    R"(
 COPY (SELECT %s FROM "%s"."%s" WHERE ctid BETWEEN '(%d,0)'::tid AND '(%d,0)'::tid %s) TO STDOUT (FORMAT binary);
```
stevefan1999-personal, line 414: but you will have to add AND back here:
```
COPY (SELECT %s FROM "%s"."%s" WHERE ctid BETWEEN '(%d,0)'::tid AND '(%d,0)'::tid AND %s) TO STDOUT (FORMAT binary);
```

postgres_scanner.cpp
FranckPachot: We need to keep WHERE only if there's a filter. I'll add that

postgres_scanner.cpp
FranckPachot: We need to add a AND only if there's a filter. I'll add that

Mytherin:
Thanks for the PR! I've removed the unnecessary `WHERE` clause in #111 as well in case of a full table scan

PR #86, opened 2023-03-08
Fix enum array support

gpsamson:
This pull request fixes an issue where enum array columns would fail with because of an empty namespace value:

```
07:23:45     ERROR:  syntax error at or near "."
07:23:45    LINE 1: SELECT unnest(enum_range(NULL::.my_enum_type))
```

For some reason, there is a condition that throws an error if an array value's length doesn't match its `typlen` value. This doesn't make sense since Postgres arrays are variable length. Existing array tests seem to pass without it.

## Testing

I added some additional unit tests for enum arrays. Github Actions was slow to start so pasting test output from my machine:

```
[100%] Built target postgres_scanner_loadable_extension
./build/release/test/unittest --test-dir . "[postgres_scanner]"
Filters: [postgres_scanner]
[17/17] (100%): test/postgres_scanner/bug76.test
===============================================================================
All tests passed (30776 assertions in 17 test cases)
```

### DIFF
--- a/postgres_scanner.cpp
+++ b/postgres_scanner.cpp
@@ -293,6 +293,7 @@ ORDER BY attnum;
 		info.type_info.typtype = res->GetString(row, 5);
 		info.typelem = res->GetInt64(row, 6);
 
+		info.elem_info.nspname = res->GetString(row, 2);
 		info.elem_info.typname = res->GetString(row, 7);
 		info.elem_info.typlen = res->GetInt64(row, 8);
 		info.elem_info.typtype = res->GetString(row, 9);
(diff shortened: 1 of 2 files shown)


gpsamson:
@Mytherin Can't request you as a reviewew, no rush to get this reviewed. Once merged in looks like I need to open a PR to bump the extensions CSV here: https://github.com/duckdb/duckdb/blob/0e959820c53bb072b290b984db1a54c2c14208b9/.github/config/extensions.csv#L12. Lmk if not. Thanks again

Mytherin:
Could we add a test file that reads from the enum tables/column as well?

Mytherin:
As far as updating the extensions in the main repo - that is indeed required to update the extension for the next release, but it won’t update the extension for the current release so there is no immediate rush.

jhrcek:
@gpsamson awesome, thanks for working on this!
I built and tested your PR locally and I'm now able to run postgres_attach without errors.

Mytherin:
Thanks!

PR #99, opened 2023-06-20
Expose pages_per_task as config parameter in postgres_scan

damian-petta:
The Postgres scanner works really well, but certain tables, in particular, demonstrate long query times, taking about an hour to load the full table. Some tables of similar size, on the other hand, work much faster. Therefore, we started investigating to find the cause of the slow ingestion from Postgres. It turned out that this was caused by a parameter that controls the number of rows per page returned in parallel during data downloading. After adjusting the 'pages_per_task' parameter from the default value of 1000 to 1,000,000, the average time reduced to 467.30 seconds from the initial one hour. That's why I believe that 'pages_per_task' should be exposed as a config parameter, and that's my proposition for the issue. (Tbh, I think it would be better if the parameter were optional, but I don't have any idea how to manage this without extra modules or bumping the C++ version to C++17)


Mytherin:
Thanks for the PR! I've added the `pg_pages_per_task` option in #111 that should address this issue as well

PR #103, opened 2023-08-02
Update ext build system

samansmink:
Update duckdb and the builds system + ci for the postgres extension.

1 casualty here is the tpcds test, which failed on manylinux with a out of storage error, I left that one to a future PR for which i will open an issue after this PR gets merged.


Mytherin:
Thanks! Looks great

PR #107, opened 2023-09-27
avoid overflow for unbounded numeric with hugeints

Alex-Monahan:
A MotherDuck design partner encountered a numeric overflow when trying to read in unbounded numeric columns from Postgres. I changed to using hugeint's to support larger numbers and added a test case.

Please let me know if you'd like to see anything changed or improved with this! I am a C++ rookie after all...


Mytherin:
Thanks for the PR! I've pushed a fix for this problem in #111 as well

PR #111, opened 2023-10-17
Add support for attaching Postgres databases as a pluggable storage catalog, and fix many issues

### LINKED ISSUE #16: Can't connect - Error: Decimal with a width of 255 is bigger than the maximum supported of 38.
unfortunately, I couldn't connect to any of my postgresql databases through the duckdb scanner due to 2 types of errors.

first one: **Decimal** with a width of 255 is bigger than the maximum supported of 38.
**I tried with both mechanisms**, to connect to the database and to connect to just a table.
`call postgres_attach('dbname=bassam');`
`select * from postgres_scan('dbname=bassam', 'public', 'test');`
Screen Shot 2022-09-30 at 10 50 55 PM

Screen Shot 2022-09-30 at 10 50 55 PM

the only database that I could able to connect to is the one that having 0 tables.

the second problem: is that the database has not been invalidated!
`D select * from postgres_scan('dbname=bassam', 'public', 'untitled_table');`
Screen Shot 2022-09-30 at 10 54 14 PM

you can replicate the issue with the this dataset below:
https://github.com/bassamsdata/earthquake_data/blob/main/earthquake.csv

postgresql version: 14.5
duckdb version: 0.5.1
macOS: 12.6

thanks a lot for your efforts.

### LINKED ISSUE #68: Filtering by char column doesn't work correctly in Postgres_scanner
### What happens?

When filtering by a char column the query doesn't return correct results if the filter is not pushed down to Postgres. For example in the ``customer`` table of TPC-H, the c_mktsegment is usually a char(10).

```select count(*), c_mktsegment from customer where c_mktsegment = 'BUILDING' group by c_mktsegment;``` 

doesn't return any results when the filter is not pushed down to Postgres. However when the filter is pushed down to Postgres, the query returns the correct result. I believe its because Postgres ignores the blank space padding of the column but duckdb does not.

Similarly, if the query is written explicitly with the blank spaces padding, and the filter isn't pushed down, then this also works OK.

```select count(*), c_mktsegment from customer where c_mktsegment = 'BUILDING  ' group by c_mktsegment;```

### To Reproduce

* Create the TPC-H ``customer`` table in **postgres** and populate it with data of any scale. Note that the  c_mktsegment is a char(10)
```
create table tpch.customer
(
    c_custkey    integer        not null,
    c_name       varchar(25)    not null,
    c_address    varchar(40)    not null,
    c_nationkey  integer        not null,
    c_phone      char(15)       not null,
    c_acctbal    numeric(15, 2) not null,
    c_mktsegment char(10)       not null,
    c_comment    varchar(117)   not null
);
```
* Link duckdb to the postgres schema and don't enable filter pushdown

``` CALL postgres_attach('host=localhost port=28815 dbname=tpch', source_schema='tpch', filter_pushdown=false, overwrite=true);```

* Run the query below, you'll see that no results are returned
```select count(*), c_mktsegment from customer where c_mktsegment = 'BUILDING' group by c_mktsegment;```

### OS:

MacOS

### PostgreSQL Version:

15

### DuckDB Version:

0.6.1

### DuckDB Client:

CLI

### Full Name:

Olo Sawyerr

### Affiliation:

None. Hobby projects.

### Have you tried this on the latest `master` branch?

- [x] I agree

- [X] I agree

### LINKED ISSUE #69: Floating point exception (core dumped)
### What happens?

```
Floating point exception (core dumped)
```

### To Reproduce

In Postgres:
```sql
create table fail(n numeric(12,7));
insert into fail values (32.8875000);
```

In DuckDB CLI:
```sql
INSTALL postgres;
LOAD postgres;
CALL postgres_attach('...')
select * from fail;
```

### OS:

Ubuntu 20.04.5 LTS

### PostgreSQL Version:

15.1

### DuckDB Version:

v0.6.1-dev83 dfae126ebb

### DuckDB Client:

CLI

### Full Name:

Dave Tapley

### Affiliation:

JE Fuller

### Have you tried this on the latest `master` branch?

- [X] I agree

- [X] I agree

### LINKED ISSUE #73: Write functionality
This is a “far out there” feature request that I suspect is not even possible to implement (it’s a scanner not a writer after all), but here it goes anyway. Any chance DuckDB could grow Postgres writing functionality?

We have data pipelines where we move data from an analytical database (Redshift and BigQuery) to Postgres. We currently do this by dumping CSVs which we then load into Postgres. This is a pain most of the time and a nightmare when you start running into delimiters in your text columns, nulls vs empty characters, etc. It would be really nice to be able to go from Parquet to Postgres (or even OLAP DB to Postgres) using the simplicity of DuckDB’s ability to scan, buffer and transform data.

Just a thought after a night of hair pulling 😅

### LINKED ISSUE #79: Binder Error: table "postgres_scan" has duplicate column name
### What happens?

I am trying to use `postgres_scan` over a heterogeneous database, but I get the following error: `Binder Error: table "postgres_scan" has duplicate column name "ID"`.

I have no idea about what that means, because there are no duplicate columns in Postgres in the same table, but the column `ID` is used across 23 tables. Some schemas have multiple `ID` columns across different tables:

```
gisdata=# select table_schema, count(*) from information_schema.columns where column_name = 'ID' group by table_schema having count(*) > 1;
   table_schema   | count 
------------------+-------
 Hydrography      |     2
 merlin_radar     |     6
 openness_tree_db |     2
 public           |     3
 Topography       |     2
(5 rows)
```

I will be happy to share more information about the database, but the database is not available on the internet, and sharing a huge database dump seems quite useless.

### To Reproduce

```
install postgres;
load postgres;
call postgres_attach('');
Binder Error: table "postgres_scan" has duplicate column name "ID"
```

### OS:

Ubuntu 20.04 x64

### PostgreSQL Version:

13.9

### DuckDB Version:

0.7.0

### DuckDB Client:

Native

### Full Name:

Francesco Frassinelli

### Affiliation:

Norwegian Institute for Nature Research

### Have you tried this on the latest `master` branch?

- [X] I agree

- [X] I agree

### LINKED ISSUE #81: Support for querying views
hello, i wanted to open this issue to track potential efforts on using `postgres_attach`/`postgres_scanner` to be able to query views. i didn't realize querying views wasn't possible until I saw it commented on #20 and i didn't see another open issue specifically for this use case.

### LINKED ISSUE #82: postgres_attach fails with SQL syntax error when postgres contains table with column of type "array of enum"
### What happens?

When you have table in postgres, containing a column whose type is "array of enum", 
running postgres_attach fails with error like this:

```
Error: IO Error: Unable to query Postgres: ERROR:  syntax error at or near "."
LINE 1: SELECT unnest(enum_range(NULL::.myenum))
                                       ^
```

### To Reproduce

1. Start latest version (15.2) of postgresql locally in docker (but it's also reproducible with older postgres, like 12.11)
```bash
docker run \
-p 5432:5432 \
-e POSTGRES_USER=test \
-e POSTGRES_PASSWORD=test \
-e POSTGRES_DB=test \
-e PGDATA=/var/lib/postgresql/data/pgdata \
--tmpfs /var/lib/postgresql/data/pgdata \
postgres:15.2-alpine \
-c log_statement=all
```
2. Connect to test db using psql
```bash
psql "host=localhost user=test password=test"
```
and init simple DB schema using following statements (note the column whose type is array of some enum)

```sql
CREATE TYPE myenum as ENUM ('whatever');
CREATE TABLE test (array_of_myenum myenum[]);
```
3. Start duckdb from CLI and try postgres_attach

```bash
~/Tmp> duckdb 
v0.7.0 f7827396d7
Enter ".help" for usage hints.
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
D LOAD postgres;
D CALL postgres_attach('host=localhost user=test password=test');
Error: IO Error: Unable to query Postgres: ERROR:  syntax error at or near "."
LINE 1: SELECT unnest(enum_range(NULL::.myenum))
                                       ^
 ERROR:  syntax error at or near "."
LINE 1: SELECT unnest(enum_range(NULL::.myenum))
                                       ^
```

### OS:

Linux - Fedora 36

### PostgreSQL Version:

reproduced in 12.11 and 15.2

### DuckDB Version:

v0.7.0 f7827396d7

### DuckDB Client:

duckdb cli

### Full Name:

Jan Hrček

### Affiliation:

Holmusk

### Have you tried this on the latest `master` branch?

- [X] I agree

- [X] I agree

### LINKED ISSUE #85: too many clients connected to PostgreSQL when using aggregate function
### What happens?

When I use the **aggregate function** (like **count(), sum()**) to access a PostgreSQL table, the postgres scanner produces a lot of PostgreSQL Clients connections.  Other SQLs are OK.
The IOException is :`duckdb.IOException: IO Error: Unable to connect to Postgres at user=xxx dbname=xxx: connection to server on socket "/tmp/.s.PGSQL.5432" failed: FATAL:  sorry, too many clients already`.

### To Reproduce

I compiled the whole project with `BUILD_PYTHON=1 make -j debug` and load the extension at **debug directory** and install  DuckDB from `duckdb/tools/pythonpkg/setup.py`.
(The `max_connections` parameter is `100` in $PGDATA/postgresql.conf.)
Here's my python code to reproduce the bug (sorry for the absolute path):
```python
import duckdb
print(duckdb.__version__) # '0.7.2-dev327'
con = duckdb.connect(":memory:", config={"allow_unsigned_extensions" : True})
con.execute("LOAD '/home/xinning/duckdb_scanner_candebug/postgres_scanner/build/debug/extension/postgres_scanner/postgres_scanner.duckdb_extension' ")
con.execute("SELECT count(*) FROM POSTGRES_SCAN('user=xinning dbname=tpch_sf1', 'public', 'lineitem') as l where l.l_orderkey > 6666;") # IO Error
con.execute("SELECT count(l.l_orderkey) FROM POSTGRES_SCAN('user=xinning dbname=tpch_sf1', 'public', 'lineitem') as l where l.l_orderkey > 6666;")  # IO Error
con.execute("SELECT sum(l.l_orderkey) FROM POSTGRES_SCAN('user=xinning dbname=tpch_sf1', 'public', 'lineitem') as l where l.l_orderkey > 6666;") # IO Error
con.execute("SELECT l.l_orderkey FROM POSTGRES_SCAN('user=xinning dbname=tpch_sf1', 'public', 'lineitem') as l where l.l_orderkey > 6666;") # OK
```

### OS:

CentOS7

### PostgreSQL Version:

15.2

### DuckDB Version:

0.7.2-dev327

### DuckDB Client:

Python

### Full Name:

Xinning Zhang

### Affiliation:

Tsinghua University

### Have you tried this on the latest `master` branch?

- [X] I agree

- [X] I agree

### LINKED ISSUE #92: Postgres NUMERIC columns - overflowing values when inserting into duckdb's DECIMAL
### What happens?

When importing data from postgres into duckdb using `postgres_scan` function, numeric values with large magnitude are overflowing in duckdb.

### To Reproduce

1. Start latest version (15.2) of postgresql locally in docker (but it's also reproducible with older postgres, like 12.11)
```bash
$ docker run \
-p 5432:5432 \
-e POSTGRES_USER=test \
-e POSTGRES_PASSWORD=test \
-e POSTGRES_DB=test \
-e PGDATA=/var/lib/postgresql/data/pgdata \
--tmpfs /var/lib/postgresql/data/pgdata \
postgres:15.2-alpine \
-c log_statement=all
```

2. Connect to test db using psql
```bash
psql "host=localhost user=test password=test"
```

Create table with NUMERIC column (from [postgres manual](https://www.postgresql.org/docs/14/datatype-numeric.html#DATATYPE-NUMERIC-DECIMAL): "Specifying NUMERIC without any precision or scale creates an “unconstrained numeric” column in which numeric values of any length can be stored, up to the implementation limits.)

```sql
create table test (mycolumn numeric);
```

Insert bunch of values of different magnitudes.

```sql
INSERT INTO test (mycolumn)
VALUES
(0.001),
(0.01),
(0.1),
(1),
(10),
(100),
(1000),
(10000),
(100000),
(1000000),
(10000000),
(100000000),
(1000000000),
(10000000000),
(100000000000),
(1000000000000),
(10000000000000),
(100000000000000),
(1000000000000000),
(10000000000000000),
(100000000000000000),
(1000000000000000000),
(10000000000000000000),
(100000000000000000000);
```

2. Import the data into duckdb.
   When you create column of type `numeric` in duckdb, it is implicitly translated to `decimal(18,3)`.
   To allow numbers up to 100000000000000000000 (1 and 20 zeros, with 3 decimal digits) I use `decimal(24,3)`.
   But this issue is reproducible also when I use the maximum allowed scale (`decimal(38,3)`).
   
```bash
$ duckdb demo.db
v0.7.1 b00b93f0b1
Enter ".help" for usage hints.
D LOAD postgres;
D create table test(mycolumn decimal(24,3));
D insert into test(mycolumn) select * from postgres_scan('host=localhost user=test password=test', 'public', 'test');
```

3. observe that the largest values are not correct:

```
D select * from test;
┌──────────────────────────┐
│         mycolumn         │
│      decimal(24,3)       │
├──────────────────────────┤
│                    0.001 │
│                    0.010 │
│                    0.100 │
│                    1.000 │
│                   10.000 │
│                  100.000 │
│                 1000.000 │
│                10000.000 │
│               100000.000 │
│              1000000.000 │
│             10000000.000 │
│            100000000.000 │
│           1000000000.000 │
│          10000000000.000 │
│         100000000000.000 │
│        1000000000000.000 │
│       10000000000000.000 │
│      100000000000000.000 │
│     1000000000000000.000 │
│    10000000000000000.000 │
│   100000000000000000.000 │
│  1000000000000000000.000 │
│ -8446744073709551616.000 │
│  7766279631452241920.000 │
├──────────────────────────┤
│         24 rows          │
└──────────────────────────┘
```

Interestingly  when I insert those two overflowing values directly in duckdb, they ARE inserted correctly:
```bash
D insert into test(mycolumn) values (10000000000000000000.000), (100000000000000000000.000);
D select * from test;
┌───────────────────────────┐
│         mycolumn          │
│       decimal(24,3)       │
├───────────────────────────┤
│                     0.001 │
│                     0.010 │
│                     0.100 │
│                     1.000 │
│                    10.000 │
...
│     10000000000000000.000 │
│    100000000000000000.000 │
│   1000000000000000000.000 │
│  -8446744073709551616.000 │ <─┤ two largest values imported from postgres
│   7766279631452241920.000 │ <─┤
│  10000000000000000000.000 │     <─┤ two largest values inserted directly from duckdb repl 
│ 100000000000000000000.000 │     <─┤
├───────────────────────────┤
│          26 rows          │
└───────────────────────────┘
```

### OS:

Linux - Fedora 36

### PostgreSQL Version:

15.2

### DuckDB Version:

v0.7.1 b00b93f0b1

### DuckDB Client:

CLI

### Full Name:

Jan Hrček

### Affiliation:

Holmusk

### Have you tried this on the latest `master` branch?

- [X] I agree

- [X] I agree

### LINKED ISSUE #93: Support for array of macaddr
### What happens?

Hello,

I have a postgresql table with an array column of mac addresses. I can't read the table from duckdb. I have the following error: `Error: INTERNAL Error: Unsupported type for ColumnDataCollection::GetCopyFunction`

It seems that macaddr is not supported [here](https://github.com/duckdblabs/postgres_scanner/blob/dbb255340da3b177ccc1b717adbe72b83ea36f13/postgres_scanner.cpp#L155). It would fine to have a conversion to varchar.

### To Reproduce

In postgresql:
```sql
CREATE TABLE test
(
    mac_addresses macaddr[]
);

TRUNCATE test;
INSERT INTO test (mac_addresses)
VALUES ( '{00:00:00:00:00:00, 00:00:00:00:00:01, 00:00:00:00:00:02}'::macaddr[] );
```

In duckdb:
```sql
INSTALL
postgres_scanner;
    LOAD
postgres_scanner;

CALL postgres_attach('***);

SELECT * FROM test;
-- gives Error: INTERNAL Error: Unsupported type for ColumnDataCollection::GetCopyFunction
```

### OS:

macOS

### PostgreSQL Version:

15

### DuckDB Version:

0.7.1

### DuckDB Client:

CLI

### Full Name:

Kamil Caglar

### Affiliation:

Serenicity

### Have you tried this on the latest `master` branch?

- [X] I agree

- [X] I agree

### LINKED ISSUE #94: Postgres_scanner does not detect the correct number of dimensions of an array in postgresql
### What happens?

if the array in postgresql in TEXT / float / VARCHAR that has two or more dimensions duckdb will either respond with either 

with '\0' on the first tuple response of the query , on the second  tuple will core dump / seg fault or go into unstable condition depending on the number of nested arrays.

### To Reproduce

any array within Postgresql with 2 or more dimensions will crash duckdb

### OS:

ubuntu 22.04

### PostgreSQL Version:

15.2   (Ubuntu 15.2-1.pgdg22.04+1)

### DuckDB Version:

0.7.1

### DuckDB Client:

CLI on unix or C++ lib.

### Full Name:

James Hackett

### Affiliation:

none

### Have you tried this on the latest `master` branch?

- [X] I agree

- [X] I agree

### LINKED ISSUE #96: PG_ATTACH takes very long to execute
### What happens?

The pg_attach method takes quite long to execute. is there any way to speed that up?

### To Reproduce

i am using pg_attach on a patroni replica on a server that has many DBs, but only attaching to a specific one. The query sometimes even a minute or more to return.

i have perhaps 20 or 30 tables in this DB, but with really few 100 of rows per table.

Is there any way i can debug / speed up the process ?

### OS:

linux

### PostgreSQL Version:

14

### DuckDB Version:

0.7.1

### DuckDB Client:

python

### Full Name:

check other issues

### Affiliation:

check other issues

### Have you tried this on the latest `master` branch?

- [X] I agree

- [X] I agree

### LINKED ISSUE #100: const pages_per_task parameter cause long query times on certain tables
### What happens?

The Postgres scanner works really well, but certain tables, in particular, demonstrate long query times, taking about an hour to load the full table. Some tables of similar size, on the other hand, work much faster. Therefore, we started investigating to find the cause of the slow ingestion from Postgres. It turned out that this was caused by a parameter that controls the number of rows per page returned in parallel during data downloading. After adjusting the 'pages_per_task' parameter from the default value of 1000 to 1,000,000, the average time reduced to 467.30 seconds from the initial one hour. That's why I believe that 'pages_per_task' should be exposed as a config parameter I think it would the better if the parameter will be optional. I've made a PR with solution where parameter is required (since I don't know if actual c++14 must stay or not and I don't have any idea how to manage this without extra modules or bumping the C++ version to C++17)

### To Reproduce

To reproduce we need a table that contain very large amount of rows and columns.

### OS:

any
### PostgreSQL Version:

any

### DuckDB Version:

any

### DuckDB Client:

any

### Full Name:

Damian Petta

### Affiliation:

.

### Have you tried this on the latest `master` branch?

- [X] I agree

- [X] I agree

### LINKED ISSUE #102: Special handling for timestamp and timestamptz 'infinity'
### What happens?

When using the postgres scanner on a tuple that contains an 'infinity' value in a TIMESTAMPTZ column, the value returned is "290273-12-10 (BC) 06:11:13.224191+10:12" (Not sure why my timezone is +10:12). The same result happens when setting duckdb TimeZone to UTC, though obviously with a different displayed value.

Postgres implementation: https://github.com/postgres/postgres/blob/master/src/include/datatype/timestamp.h#L150C28-L150C40
Duckdb implementation: https://github.com/duckdb/duckdb/blob/master/src/include/duckdb/common/types/timestamp.hpp#L71

### To Reproduce

On postgres (reproduced locally and on remote postgres server running postgres 14.5 on rds graviton)
```sql
create table thing_state (
thing_id int not null,
valid_from timestamptz not null,
valid_till timestamptz not null,
unique(thing_id, valid_from)
);
insert into thing_state (thing_id, valid_from, valid_till) select 1, now(), 'infinity'::timestamptz;
```

In duckdb:
```sql
load postgres;
create schema pg;
call postgres_attach('dbname=test', sink_schema=pg);
select * from pg.thing_state;
```

### OS:

MacOS 13.4.1

### PostgreSQL Version:

14.8

### DuckDB Version:

0.8.1

### DuckDB Client:

CLI

### Full Name:

Wren

### Affiliation:

Australian Greens

### Have you tried this on the latest `master` branch?

- [X] I agree

- [X] I agree

Mytherin:
This is a giant PR that reworks most of the extension.

### Pluggable Storage
This PR enables attaching Postgres databases as a pluggable catalog using the `ATTACH` command, similar to what is possible for SQLite databases (see duckdb/duckdb#6066).

Here is an example of how this works:

```sql
ATTACH 'dbname=postgresscanner' AS postgres (TYPE POSTGRES);
CREATE TABLE postgres.pg_tbl(id INT);
INSERT INTO postgres.pg_tbl VALUES (42), (84), (NULL);
SELECT * FROM postgres.pg_tbl;
┌───────┐
│  id   │
│ int32 │
├───────┤
│    42 │
│    84 │
│  NULL │
└───────┘
```

The following operations are supported:

###### CREATE TABLE
```sql
CREATE TABLE postgres.pg_tbl(id INT);
```

###### INSERT INTO
```sql
INSERT INTO postgres.pg_tbl VALUES (42), (84), (NULL);
```

###### COPY
```sql
COPY (SELECT 1) TO 'file.parquet';
COPY postgres.pg_tbl FROM 'file.parquet';
```

###### SELECT
```sql
SELECT * FROM postgres.pg_tbl;
┌───────┐
│  id   │
│ int32 │
├───────┤
│    42 │
│    84 │
│  NULL │
│     1 │
└───────┘
```

###### DESCRIBE
```sql
DESCRIBE postgres.pg_tbl;
┌─────────────┬─────────────┬─────────┬─────────┬─────────┬───────┐
│ column_name │ column_type │  null   │   key   │ default │ extra │
│   varchar   │   varchar   │ varchar │ varchar │ varchar │ int32 │
├─────────────┼─────────────┼─────────┼─────────┼─────────┼───────┤
│ id          │ INTEGER     │ YES     │ NULL    │ NULL    │  NULL │
└─────────────┴─────────────┴─────────┴─────────┴─────────┴───────┘
```

###### DELETE
```sql
DELETE FROM postgres.pg_tbl WHERE id=1;
SELECT * FROM postgres.pg_tbl;
┌───────┐
│  id   │
│ int32 │
├───────┤
│    42 │
│    84 │
│  NULL │
└───────┘
```

###### UPDATE
```sql
UPDATE postgres.pg_tbl SET id=id+1;
SELECT * FROM postgres.pg_tbl;
┌───────┐
│  id   │
│ int32 │
├───────┤
│    43 │
│    85 │
│  NULL │
└───────┘
```

###### Transactions (BEGIN, COMMIT, ROLLBACK)

```sql
BEGIN;
DELETE FROM postgres.pg_tbl;
FROM postgres.pg_tbl;
┌────────┐
│   id   │
│ int32  │
├────────┤
│ 0 rows │
└────────┘
ROLLBACK;
FROM postgres.pg_tbl;
┌───────┐
│  id   │
│ int32 │
├───────┤
│    43 │
│    85 │
│  NULL │
└───────┘
```

###### ALTER TABLE

```sql
ALTER TABLE postgres.pg_tbl ADD COLUMN name VARCHAR;
FROM postgres.pg_tbl;
┌───────┬─────────┐
│  id   │  name   │
│ int32 │ varchar │
├───────┼─────────┤
│    43 │ NULL    │
│    85 │ NULL    │
│  NULL │ NULL    │
└───────┴─────────┘
```

###### DROP
```sql
DROP TABLE postgres.pg_tbl;
FROM postgres.pg_tbl;
-- Table with name pg_tbl does not exist!
```

###### CREATE SCHEMA
```sql
CREATE SCHEMA postgres.new_schema;
CREATE TABLE postgres.new_schema.new_tbl(i INT);
INSERT INTO postgres.new_schema.new_tbl VALUES (42);
SELECT * FROM postgres.new_schema.new_tbl;
┌───────┐
│   i   │
│ int32 │
├───────┤
│    42 │
└───────┘
DROP SCHEMA postgres.new_schema CASCADE;
```

###### CREATE VIEW
```sql
CREATE VIEW postgres.v1 AS SELECT 42;
FROM postgres.v1;
┌───────┐
│  42   │
│ int32 │
├───────┤
│    42 │
└───────┘
```

###### CREATE INDEX
```sql
CREATE TABLE postgres.pg_tbl(id INT);
CREATE INDEX i_index ON postgres.pg_tbl(id);
```

###### CREATE TYPE
```sql
USE postgres;
CREATE TYPE my_point AS STRUCT(x INT, y INT);
CREATE TABLE table_with_point(point MY_POINT);
INSERT INTO table_with_point VALUES ({'x': 42, 'y': 84});
FROM table_with_point;
┌──────────────────────────────┐
│            point             │
│ struct(x integer, y integer) │
├──────────────────────────────┤
│ {'x': 42, 'y': 84}           │
└──────────────────────────────┘
```

### Metadata caching

The metadata (e.g. which schemas are there, which tables are there, what their structure is, etc) is read once and then cached. That means that if e.g. a new table is created through a different connection, the cached metadata will not know about the existence of this table. There is a new function `pg_clear_cache` that can be used to clear the metadata cache and force new data to be read from the Postgres instance.

```sql
CALL pg_clear_cache();
```

### Extra Parameters/Configuration
This PR adds the following settings to the extension:

```sql
D FROM duckdb_settings() WHERE name LIKE 'pg_%';
┌─────────────────────────────────┬─────────┬────────────────────────────────────────────────────────────────────────────┬────────────┐
│              name               │  value  │                                description                                 │ input_type │
│             varchar             │ varchar │                                  varchar                                   │  varchar   │
├─────────────────────────────────┼─────────┼────────────────────────────────────────────────────────────────────────────┼────────────┤
│ pg_debug_show_queries           │ false   │ DEBUG SETTING: print all queries sent to Postgres to stdout                │ BOOLEAN    │
│ pg_experimental_filter_pushdown │ false   │ Whether or not to use filter pushdown (currently experimental)             │ BOOLEAN    │
│ pg_connection_limit             │ 64      │ The maximum amount of concurrent Postgres connections                      │ UBIGINT    │
│ pg_array_as_varchar             │ false   │ Read Postgres arrays as varchar - enables reading mixed dimensional arrays │ BOOLEAN    │
│ pg_pages_per_task               │ 1000    │ The amount of pages per task                                               │ UBIGINT    │
│ pg_use_binary_copy              │ true    │ Whether or not to use BINARY copy to read data                             │ BOOLEAN    │
└─────────────────────────────────┴─────────┴────────────────────────────────────────────────────────────────────────────┴────────────┘
```

### Fixes

This PR reworks the extension significantly - and also fixes many outstanding issues.

Fixes #16
Fixes #68
Fixes #69
Fixes #73
Fixes #79
Fixes #81
Fixes #82
Fixes #85
Fixes #92
Fixes #93
Fixes #94
Fixes #96
Fixes #100
Fixes #102
Supersedes #59
Supersedes #99
Supersedes #107

#### Decimals

There are a number of fixes for decimals, in particular:

* Large decimals are read as doubles (e.g. when width > 38 which is the max width DuckDB itself supports)
* Large decimals are read directly into doubles 
* Overflows when reading decimals are fixed and now correctly checked for by greatly expanding the decimals that are tested

#### Support for querying views

Views can now be queried, e.g.:

###### Postgres
```sql
CREATE VIEW v1 AS SELECT 42;
```
###### DuckDB
```sql
ATTACH 'dbname=test' AS postgres (TYPE POSTGRES);
SELECT * FROM postgres.v1;
┌──────────┐
│ ?column? │
│  int32   │
├──────────┤
│       42 │
└──────────┘
```

#### Duplicate column/table names

In Postgres, column and table names are case sensitive. Hence this is valid SQL:

```sql
CREATE TABLE mixed_names("COL" INT, "col" INT);
```

This conflicts with DuckDB which is case insensitive. This PR fixes the issue by instead renaming the columns when read into DuckDB instead of throwing an error. For example:

```sql
D SELECT * FROM POSTGRES_SCAN('dbname=test', 'public', 'mixed_names');
┌───────┬───────┐
│  COL  │ col:1 │
│ int32 │ int32 │
├───────────────┤
│    0 rows     │
└───────────────┘
```

#### Connection Pool

This PR introduces the concept of a "connection pool" which is a limit to how many parallel connections will be opened during scans. The default limit is set to 64, and can be adjusted using the `pg_connection_limit` setting (e.g. `SET pg_connection_limit=1000`).

#### Complex Types

This PR adds support for complex types, including composite types (created through `CREATE TYPE`), and multidimensional arrays. This support is for both ingesting data (using e.g. CREATE TABLE/INSERT) as well as for reading data.

#### ATTACH time

While this PR does not fix the issue of `pg_attach` taking a long time to execute, the new `ATTACH` method (which should be the preferred method of attaching going forward) should not suffer from the same problem because it only issues a single query to get the column names/types of all tables at once instead of sending one query per table.


Nintorac:
Hey, great effort!!

Is there an ETA on when this will make it into a DuckDB release?

Mytherin:
Should be in next week for 0.9.2

tmontes:
Perfect timing! ...in the process of leading the first duckdb/PostgreSQL integration project.

Thanks a lot!

PR #116, opened 2023-10-23
add reusable workflows for deploying nightly binaries

samansmink:
This PR adds the reusable workflow for extension deployment to the postgress scanner, which allows producing nightly binaries. In the process, I removed the workflows that would be duplicates, leaving only the linux job that would actually run the tests with the postgres server. Also i added a require-env to all tests because the reusable workflow will call `make test`

@Mytherin Before merging we need to add the s3 credentials and signing token to the repo. Maybe we can add them as an [organisation secret](https://docs.github.com/en/codespaces/managing-codespaces-for-your-organization/managing-secrets-for-your-repository-and-organization-for-github-codespaces#adding-secrets-for-an-organization)?


.github/workflows/MainDistributionPipeline.yml
```diff
@@ -0,0 +1,31 @@
+#
+# This workflow calls the main distribution pipeline from DuckDB to build, test and (optionally) release the extension
+#
+name: Main Extension Distribution Pipeline
+on:
+  push:
+  pull_request:
```
Mytherin, line 7: Should this run on pull requests as well?

Mytherin (commented):
Thanks for the PR! Looks great. One minor comment. For the organization secrets, perhaps @hannes can have a look?

samansmink:
> Thanks for the PR! Looks great. One minor comment. For the organization secrets, perhaps @hannes can have a look?

I'll chat with hannes today on this

.github/workflows/MainDistributionPipeline.yml
samansmink: I think so? I removed most of the other jobs so without it, we would only test on linux gcc4 for PRs. I could add a `needs` to make it depend on the `linux-tests-postgres` to reduce ci time for broken builds?

Mytherin:
Thanks for the updates - is this ready to merge?

samansmink:
yea think so!

PR #124, opened 2023-11-02
Improve curl in pgconfigure to add some logging and retries

carlopi:
Should fix some random failures in fetching postgres that seems to be random and network related.

Fixes https://github.com/duckdblabs/duckdb-internal/issues/597


carlopi:
I have no real way of testing whether this makes main CI more stable, but should not be problematic to add this.

Mytherin:
Thanks!

PR #180, opened 2024-02-09
[Dev] Apply upstream patches

carlopi:


.github/workflows/Linux.yml
samansmink, line 1: This one should stay: it's used to test against a postgres server

samansmink (commented):
1 small comment, the rest looks good!

.github/workflows/Linux.yml
carlopi: Thanks! Removing Linux.yml was a bit too far, thanks for having another look.

carlopi:
I think this should be ready to go in.
Failure on the Linux test is due to the fact that this was not properly tested yet after changes to main forbidding the `"MyColumn:1"` syntax (should be probably changed to `"tbl_with_case_sensitive_columns.MyColumn"`)

Mytherin:
Yolo

PR #183, opened 2024-02-20
Add PostgresViewEntry

Tishj:
In the table query we get both tables, views, materialized views, partitioned tables and foreign tables.
The views are then stored in the catalog as a TableCatalogEntry, which confuses `duckdb_tables()` and `duckdb_views()` which is used inside `information_schema.tables`

So the views showed up as tables, this PR fixes that and adds a test for it

### DIFF
--- a/src/include/storage/postgres_view_entry.hpp
+++ b/src/include/storage/postgres_view_entry.hpp
@@ -28,8 +28,8 @@ public:
 		create_info = make_uniq<CreateViewInfo>((SchemaCatalogEntry &)schema, view);
 		// create_info->columns.SetAllowDuplicates(true);
 	}
-public:
 
+public:
 	const string &GetName() const override {
 		return create_info->view_name;
 	}
@@ -55,6 +55,7 @@ class PostgresViewEntry : public ViewCatalogEntry {
 public:
 	PostgresViewEntry(Catalog &catalog, SchemaCatalogEntry &schema, CreateViewInfo &info);
 	PostgresViewEntry(Catalog &catalog, SchemaCatalogEntry &schema, PostgresViewInfo &info);
+
 public:
 	//! Postgres type annotations
 	vector<PostgresType> postgres_types;
--- a/src/include/storage/postgres_view_entry.hpp
+++ b/src/include/storage/postgres_view_entry.hpp
@@ -1,0 +1,69 @@
+//===----------------------------------------------------------------------===//
+//                         DuckDB
+//
+// storage/postgres_view_entry.hpp
+//
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "duckdb/catalog/catalog_entry/view_catalog_entry.hpp"
+#include "duckdb/parser/parsed_data/create_view_info.hpp"
+#include "storage/postgres_create_info.hpp"
+
+namespace duckdb {
+
+struct PostgresViewInfo : public PostgresCreateInfo {
+public:
+	PostgresViewInfo() {
+		create_info = make_uniq<CreateViewInfo>();
+		// create_info->columns.SetAllowDuplicates(true);
+	}
+	PostgresViewInfo(const string &schema, const string &view) {
+		create_info = make_uniq<CreateViewInfo>(string(), schema, view);
+		// create_info->columns.SetAllowDuplicates(true);
+	}
+	PostgresViewInfo(const SchemaCatalogEntry &schema, const string &view) {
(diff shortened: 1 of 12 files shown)


src/include/storage/postgres_view_entry.hpp
```diff
@@ -0,0 +1,70 @@
...
+#include "duckdb/parser/parsed_data/create_view_info.hpp"
+#include "storage/postgres_create_info.hpp"
+
+namespace duckdb {
+
+struct PostgresViewInfo : public PostgresCreateInfo {
+public:
+	PostgresViewInfo() {
+		create_info = make_uniq<CreateViewInfo>();
+		// create_info->columns.SetAllowDuplicates(true);
```
Tishj, line 21: I'm not sure about this, the Table has a ColumnsList which provides this option, but no such option exists for views because we don't have a ColumnsList, we instead just store the types+names separately

Tishj:
This is lacking testing still, and there are probably virtual methods that need to be overridden

Tishj:
The failure is probably related to:
```sql
statement ok
CREATE VIEW s1.v1 AS SELECT 42 as j
```

Our error check is probably not a fan of this, because there is no existing table to reference

Mytherin (commented):
Can you merge with main? Also one minor comment:

Tishj:
@Mytherin This seems like a problem, I can't really find the reason for this behavior
```
# Not aliased (no expression alias)
statement ok
CREATE VIEW s1.view_aliases.x AS SELECT 'x'

query I
select x from s1.view_aliases.x
----
{'CAST('x' AS VARCHAR)': x}
```

I've had to replace the CreateViewInfo names with the expression.ToString() if the name was "?column?", this seems to have this effect

Mytherin:
The fact that we are getting two different names (one from Postgres', one from DuckDB') indicates to me we are binding the same query both in Postgres and in DuckDB. That seems like a more fundamental problem to me. After all - if a view contains anything that DuckDB does not support, the binding in DuckDB will fail, whereas we should still be able to query the view.

Maybe we could try e.g. not pulling the SQL definition of the view into DuckDB so the binding is avoided?

Mytherin:
An example of a view that works in Postgres but wouldn't work in DuckDB might be:

```sql
create table my_table(i int);
create view my_view as select ctid from my_table;
```

Tishj:
> An example of a view that works in Postgres but wouldn't work in DuckDB might be:
> 
> ```sql
> create table my_table(i int);
> create view my_view as select ctid from my_table;
> ```

You're right, that doesn't work
But that also doesn't work for `select ctid from s1.my_table`, because we've never registered `ctid` as being part of `my_table` in our custom postgres catalog

Tishj:
The problem with this lies in `Binder::Bind(BaseTableRef &ref)`
Even if I remove the query from the created ViewCatalogEntry, this code looks up the catalog entry, binds it and compares the new results versus the old results, to throw the `"Contents of view were altered: types don't match!"` error.

(also setting the query to nullptr causes an InternalException here currently because this path assumes the query is always present)
Perhaps we want to just not do this verification step if the saved query is nullptr?

Tishj:
The problem here is that we first bind the view's query before we hand it off to postgres, so we have to be able to bind it ourselves (possibly delegating to the postgres catalog extension)

Then we create the entry in the postgres catalog, running the same query, then reloading to get the entry as it was created in postgres.
This saves it as "?column?" because that's what postgres assigned to it.

Then in our verification step inside Bind(BaseTableRef &ref) we bind the query again on duckdb's side, which produces a different name for the column.

------------

The "skip view content verification if `query` is nullptr" solution could work in that case, because we won't bind the query on our side again.
That still leaves the problem of the first bind though, if system columns are referenced that we haven't explicitly made a part of the TableCatalogEntry the bind will fail (as evidenced by the `create view my_view as select ctid from my_table` example)

One solution there might be decompose the TableCatalogEntry into a catalog set containing columns, and using a DefaultGenerator to create the system columns
Another could be to always add the system columns manually in our (postgres)TableCatalogEntry creation code.

A more generic approach that will likely take the least effort is to add a virtual method on the table catalog entry for column retrieval, so we can hook in and return a column definition for "ctid" and friends, basically a poor mans "DefaultColumnGenerator"

Tishj:
We're likely going to have to separate out the SelectStatement inside the CreateViewInfo, then construct a postgres query from that, get the names + types from the PGresult object, and convert the type into a LogicalType

Issue #204, opened 2024-04-03
Support encoding to a file


adriangb:
Copying over from https://github.com/duckdb/duckdb/discussions/11438

DuckDB has become an amazing in memory ETL/data pipeline tool. It can read and write to almost anything. Would it be possible to add an option to copy to any DuckDB source to a binary postgres dump *file*? The equivalent of copying a parquet file (for example) to a postgres table then doing `\copy mytable TO 'data.bin' WITH (FORMAT BINARY)` via psql.

One option for an API would be something like `COPY mytable TO STDOUT WITH (FORMAT POSTGRES_BINARY)` / `COPY 'file.parquet' TO 'data.bin' WITH (FORMAT POSTGRES_BINARY)`. I'm hoping this is not a difficult request since it's in some ways a subset of the functionality alraedy implemented.

As an example use case I made a library for this that DuckDB could then replace and it has 200+ stars, which is more than this repo currently has (not that it's a fair comparison, just pointing out that it's a real use case): https://github.com/adriangb/pgpq

Mytherin:
Thanks - that's an interesting idea. I've opened a PR that implements this in https://github.com/duckdb/postgres_scanner/pull/205. Indeed it can utilize existing logic. Perhaps give it a spin and see if it works as you expect it to?

adriangb:
Just to complete the loop, once the next release is made this will be possible:

```python
import duckdb
from fsspec import filesystem

fs = filesystem('memory')
duckdb.register_filesystem(fs)
duckdb.sql("LOAD postgres;'")
duckdb.sql("COPY (SELECT 1 AS a, 2 AS b) TO 'memory://test.bin' (FORMAT postgres_binary)")
print(fs.open('test.bin').read())  # b'PGCOPY\n\xff\r\n\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x02\x00\x00\x00\x04\x00\x00\x00\x01\x00\x00\x00\x04\x00\x00\x00\x02\xff\xff'
```

PR #205, opened 2024-04-03
Implement #204 - Add support for POSTGRES_BINARY copy method

Mytherin:
Implements #204

This PR adds support for the `POSTGRES_BINARY` copy format, that allows you to write files in the [Postgres binary format](https://www.postgresql.org/docs/current/sql-copy.html). The files can be subsequently copied into Postgres using their `FORMAT binary` copy method.  Example use case:

###### DuckDB
```sql
D create table integers(i int);
D insert into integers from range(100);
D copy integers to 'pg_binary.bin' (format postgres_binary);
```
###### Postgres
```sql
postgres=# create table integers(i int);
postgres=# copy integers from '/{working_directory}/pg_binary.bin' (format binary);
postgres=# SELECT COUNT(*), MIN(i), MAX(i), SUM(i) FROM integers;
 count | min | max | sum  
-------+-----+-----+------
   100 |   0 |  99 | 4950
(1 row)
```

This uses the `PostgresBinaryWriter` that we have developed. Not all data types are supported - for now this is limited to types that directly map to Postgres - `bool, smallint, int, bigint, float, double, decimal, date, time, timestamp, interval, uuid, blob, varchar` and composite types (arrays/structs). We could probably create a mapping for other types (e.g. unsigned types to Postgres' signed types). 

For now this only supports writing - although we could probably support reading from the binary format using the `PostgresBinaryReader` as well.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -5,6 +5,7 @@ add_subdirectory(storage)
 add_library(
   postgres_ext_library OBJECT
   postgres_attach.cpp
+  postgres_binary_copy.cpp
   postgres_connection.cpp
   postgres_copy_from.cpp
   postgres_copy_to.cpp
--- a/src/include/postgres_binary_copy.hpp
+++ b/src/include/postgres_binary_copy.hpp
@@ -1,0 +1,36 @@
+//===----------------------------------------------------------------------===//
+//                         DuckDB
+//
+// postgres_binary_copy.hpp
+//
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "postgres_utils.hpp"
+#include "postgres_result.hpp"
+#include "duckdb/function/copy_function.hpp"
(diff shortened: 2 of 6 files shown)


adriangb:
I'm amazed at how little extra code this was and how fast you implemented it, this is really an amazing project!

I tested this and it worked as described with no issues 😄 

The one thing missing for DuckDB to replace pgpq is a way to copy to an in-memory buffer in Python at least, but that's tangential to this PR, I think this should go ahead just like this!

Mytherin:
Thanks for trying it out. Copying to an in-memory buffer might be possible with `fsspec`, although I'm not super well versed in that so I'm not entirely sure.

adriangb:
Thanks certainly not related to this so I opened https://github.com/duckdb/duckdb/discussions/11499 to continue that discussion. Thank you again for getting this done so quickly!

Issue #213, opened 2024-04-16
Add support for managing Postgres credentials through the secret manager


Mytherin:
DuckDB now includes a secret manager that can be used to manage secrets - we should add support for managing Postgres' databases through this mechanism as well.

LMnet:
Small addition - not only for postgres, but for any `attach`. There is also mysql.

Mytherin:
And now in MySQL in https://github.com/duckdb/duckdb_mysql/pull/64

PR #217, opened 2024-04-19
Allow Postgres credentials to be managed using the SecretManager

Mytherin:
Implements #213

This allows credentials to be managed using secrets. The secret to be used can be selected with the `SECRET` setting that is passed into `ATTACH`. There is also a default secret - this is the secret that is created without specifying a name. When no secret is provided to `ATTACH`, and such a secret is present, this secret is used.

Secrets can provide the following set of options.

|    Name    |             Description              |
|------------|--------------------------------------|
| `dbname`   | Database name                        |
| `database` | Alias for dbname                     |
| `user`     | Postgres user name                   |
| `password` | Postgres password                    |
| `host`     | Name of host to connect to           |
| `port`     | Port number                          |

Note that not all of these need to be provided. For options that are not provided, the system falls back to the regular set of default options as specified [in the configuration](https://duckdb.org/docs/extensions/postgres#configuration).

##### Implicit (Default) Secret With All Options
```sql
CREATE SECRET (
	TYPE POSTGRES,
	HOST '127.0.0.1',
        PORT 5432,
	DATABASE postgresscanner,
        USER 'postgres',
	PASSWORD ''
);
ATTACH '' AS secret_attach (TYPE POSTGRES);
```

##### Explicit (Named) Secret With Limited Options
```sql
CREATE SECRET postgres_secret(
	TYPE POSTGRES,
	HOST '127.0.0.1',
	DATABASE postgresscanner,
	PASSWORD ''
);

ATTACH '' AS secret_attach (TYPE POSTGRES, SECRET postgres_secret);
```

##### Overriding Options
Secrets essentially provide a new set of **default** options. The individual options can be overridden in the attach string, similar to how they can be overridden for the standard set of default options. For example:

```sql
CREATE SECRET (
	TYPE POSTGRES,
	HOST '127.0.0.1',
        PORT 5432,
	DATABASE postgresscanner,
        USER 'postgres',
	PASSWORD ''
);
-- override the dbname option, but use the rest of the options provided in the secret
ATTACH 'dbname=unknown_db_name' AS secret_attach (TYPE POSTGRES);
-- FATAL:  database "unknown_db_name" does not exist
```

### DIFF
--- a/src/postgres_extension.cpp
+++ b/src/postgres_extension.cpp
@@ -72,6 +72,44 @@ static void SetPostgresDebugQueryPrint(ClientContext &context, SetScope scope, V
 	PostgresConnection::DebugSetPrintQueries(BooleanValue::Get(parameter));
 }
 
+unique_ptr<BaseSecret> CreatePostgresSecretFunction(ClientContext &context, CreateSecretInput &input) {
+	// apply any overridden settings
+	vector<string> prefix_paths;
+	auto result = make_uniq<KeyValueSecret>(prefix_paths, "postgres", "config", input.name);
+	for (const auto &named_param : input.options) {
+		auto lower_name = StringUtil::Lower(named_param.first);
+
+		if (lower_name == "host") {
+			result->secret_map["host"] = named_param.second.ToString();
+		} else if (lower_name == "user") {
+			result->secret_map["user"] = named_param.second.ToString();
+		} else if (lower_name == "database") {
+			result->secret_map["dbname"] = named_param.second.ToString();
+		} else if (lower_name == "dbname") {
+			result->secret_map["dbname"] = named_param.second.ToString();
+		} else if (lower_name == "password") {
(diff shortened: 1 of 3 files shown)


src/postgres_extension.cpp
```diff
@@ -72,6 +72,44 @@ static void SetPostgresDebugQueryPrint(ClientContext &context, SetScope scope, V
...
+		auto lower_name = StringUtil::Lower(named_param.first);
+
+		if (lower_name == "host") {
+			result->secret_map["host"] = named_param.second.ToString();
+		} else if (lower_name == "user") {
+			result->secret_map["user"] = named_param.second.ToString();
+		} else if (lower_name == "database") {
+			result->secret_map["dbname"] = named_param.second.ToString();
+		} else if (lower_name == "dbname") {
+			result->secret_map["dbname"] = named_param.second.ToString();
```
Tishj, line 89: Do we want to throw if both `database` and `dbname` are provided?

Issue #134, opened 2023-11-19
Support Postgres on AWS Aurora


otarmelia:
### What happens?

postgres_scan function does not supports postgres in aws aurora. When executing statement: 

_SELECT * FROM postgres_scan('dbname=db user=postgres password=*** host=127.0.0.1', 'public', 'table_name');_

it gives error: 

**Error: Invalid Error: Failed to execute query "SELECT pg_is_in_recovery(), pg_export_snapshot(), (select count(*) from pg_stat_wal_receiver)": ERROR:  Function pg_stat_get_wal_receiver() is currently not supported in Aurora**

### To Reproduce

Create postgresql database in amazon RDS and execurte query: 
SELECT * FROM postgres_scan('dbname=db user=postgres password=*** host=127.0.0.1', 'public', 'table_name');

### OS:

Windows

### PostgreSQL Version:

PostgreSQL 13.9

### DuckDB Version:

v0.9.2

### DuckDB Client:

CLI

### Full Name:

Otar

### Affiliation:

Personal

### Have you tried this on the latest `main` branch?

- [X] I agree

### Have you tried the steps to reproduce? Do they include all relevant data and configuration? Does the issue you report still appear there?

- [X] I agree

Mytherin:
Thanks for the report! This looks like a duplicate of #129 which has already been fixed in #131. Could you try this out with the latest development build? If the problem persists there then feel free to re-open.

StephanGeorg:
@Mytherin 

No, does not work with latest build. Last working version is DuckDB v.0.9.1.

Mytherin:
It has been fixed in the latest dev build, not the latest release build. Which one are you using?

Mytherin:
Try using the new `ATTACH` workflow instead of `postgres_scan`/`postgres_attach`. `postgres_attach` and `postgres_scan` are deprecated.

i.e.:

```sql
ATTACH 'dbname=db user=postgres host=127.0.0.1 port=5432' AS pgdb (TYPE POSTGRES);
```

StephanGeorg:
Thanks. That works.

Issue #90, opened 2023-03-23
postgres_attach() error when connecting to read-only replica


loicalleyne:
I'm trying to connect to a postgres read-only replica and getting the following error. The replica is using log shipping to prevent replication delays from cascading up to the master DB in production.

```
call postgres_attach('host=< IP address > port=5432 user=< USER > dbname=< DB NAME > sslmode=require', source_schema='public', sink_schema='abc');
Error: IO Error: Unable to query Postgres: ERROR:  cannot assign TransactionIds during recovery
 ERROR:  cannot assign TransactionIds during recovery
```

Is this something that can be supported with the pg library being used? Connecting to the same DB from the same machine with psql works fine.

seanabraham:
Seeing the exact same issue with a postgres read replica, I don't have much to add except that the same error occurs with a `SELECT * from postgres_scan(...)` as well.

Mytherin:
Thanks for the report! This should be addressed in #112. I've tested this with a read-replica through AWS RDS but it would be great if you could verify that it resolves the issue.

seanabraham:
Amazing, thank you @Mytherin. How does this get deployed so to speak? I see that DuckDB bundles this extension and presumably this commit hash needs to be bumped: https://github.com/duckdb/duckdb/blob/5ec85a719940a9fade15c38e7601712e9cef58d8/.github/config/extensions.csv#L12

Is there a way to test this version of `postgres_scanner` locally somehow?

Mytherin:
You can compile the postgres_scanner by cloning the repo and running `make`. The extension can then be found in `build/release/extension/postgres_scanner/postgres_scanner.duckdb_extension` which can be installed using the SQL command:

```sql
FORCE INSTALL 'build/release/extension/postgres_scanner/postgres_scanner.duckdb_extension';
```

PR #235, opened 2024-05-31
Allow synchronous snapshot on replicas and in RDS/Aurora

bdavisab:
Synchronized snapshots are supported on replicas starting with PostgreSQL 10, and they're also supported in RDS/Aurora.

The change in Postgres 10 didn't make it into the release notes, but I found this https://github.com/postgres/postgres/blob/master/src/bin/pg_dump/pg_dump.c#L1343

Postgres 9.6 is super old now, even 10 and 11 aren't supported anymore, and 12 will be EOL in November.  So I think it's reasonable to simply drop the version checks.  But we can keep the guards in too if you prefer, happy to update.

Tested with an Aurora replica, and that works just fine as well.

It's true that in https://github.com/duckdb/postgres_scanner/issues/134 , pg_stat_get_wal_receiver() isn't supported in Aurora, but that's just used to check if it's a replica or not, which since PG10, we don't need to worry about.

I assume that the people reporting this issue https://github.com/duckdb/postgres_scanner/issues/90 were just using ancient versions of Postgres.  Maybe that's an argument for keeping a version check?  Let me know what you think.

### DIFF
--- a/src/postgres_scanner.cpp
+++ b/src/postgres_scanner.cpp
@@ -62,40 +62,21 @@ private:
 	PostgresConnection connection;
 };
 
-static void PostgresGetSnapshot(PostgresVersion version, const PostgresBindData &bind_data,
-                                PostgresGlobalState &gstate) {
+static void PostgresGetSnapshot(const PostgresBindData &bind_data, PostgresGlobalState &gstate) {
 	unique_ptr<PostgresResult> result;
 	// by default disable snapshotting
 	gstate.snapshot = string();
 	if (gstate.max_threads <= 1) {
 		return;
 	}
-	if (version.type_v == PostgresInstanceType::AURORA) {
-		return;
-	}
+
 	// reader threads can use the same snapshot
 	auto &con = gstate.GetConnection();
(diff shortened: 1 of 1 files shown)


JelteF:
Took a quick look at this. My understanding is that this combines two things into one PR:
1. Removing support for Postgres 9.6 (which seems reasonable to me)
2. Supporting synchronous snapshots in AWS Aurora.

I personally would prefer these two things to be done in separate PRs, but I'm not a maintainer on this repo.

jgmills98:
Hi All,

Any update on this? Re-enabling this for aurora would help drive a project I am working on.

Although I am not too familiar with this codebase, if the only blocker is splitting this into two PRs, I'd be happy to help. For re-enabling aurora, perhaps a default to the current behavior with an optional override would be the best way forward.

Warm regards

staticlibs:
Hi! I'll close this in favour of #524. It was also added to 1.5 branch, going to be published to `core_nightly` in a few minutes, to try it on 1.5.4 (or 1.5.5):

```sql
FORCE INSTALL postgres FROM core_nightly
```

PR #243, opened 2024-07-19
Bump to DuckDB 1.0.0

carlopi:


carlopi:
Given extension on v1.0.0 is already published, this failure is already there. This PR just justify / verify this is the case. To be solved independently

Mytherin:
Thanks!

PR #259, opened 2024-09-03
Fix #221: add the SCHEMA parameter to ATTACH that allows you to specify which schema to load

### LINKED ISSUE #221: New ATTACH very slow compared to postgres_attach
### What happens?

When connecting to our production database, which has hundreds of schemas, each with over a hundred tables, using the new ATTACH command takes much longer than the now deprecated `postgres_attach` function.

We've even tried passing in `--search_path=my_schema` as an option to the connection string, but didn't see any real improvement.

Having the `pg_debug_show_queries=true` and running the initial queries on a separate connection, we see that this query

```sql
SELECT pg_namespace.oid AS namespace_id, relname, relpages, attname,
    pg_type.typname type_name, atttypmod type_modifier, pg_attribute.attndims ndim,
    attnum, pg_attribute.attnotnull AS notnull, NULL constraint_id,
    NULL constraint_type, NULL constraint_key
FROM pg_class
JOIN pg_namespace ON relnamespace = pg_namespace.oid
JOIN pg_attribute ON pg_class.oid=pg_attribute.attrelid
JOIN pg_type ON atttypid=pg_type.oid
WHERE attnum > 0 AND relkind IN ('r', 'v', 'm', 'f', 'p')
UNION ALL
SELECT pg_namespace.oid AS namespace_id, relname, NULL relpages, NULL attname, NULL type_name,
    NULL type_modifier, NULL ndim, NULL attnum, NULL AS notnull,
    pg_constraint.oid AS constraint_id, contype AS constraint_type,
    conkey AS constraint_key
FROM pg_class
JOIN pg_namespace ON relnamespace = pg_namespace.oid
JOIN pg_constraint ON (pg_class.oid=pg_constraint.conrelid)
WHERE relkind IN ('r', 'v', 'm', 'f', 'p') AND contype IN ('p', 'u')
ORDER BY namespace_id, relname, attnum, constraint_id;
```
in particular takes the longest to run when attaching the database. Is there any way to optimize this query? Perhaps allowing a parameter to specify only some schemas to load?

### To Reproduce

```sql
INSTALL postgres;
LOAD postgres;
SET pg_debug_show_queries = true;
ATTACH '{self._psql_string}' AS postgres_db (TYPE POSTGRES, READ_ONLY);
```

The ATTACH statement takes about 6 minutes on average every time, compared to the previous method

```sql
INSTALL postgres;
LOAD postgres;
CALL postgres_attach(
    '{self._psql_string}',
    source_schema='{schema}',
    filter_pushdown=1,
    overwrite=1
);
```

which takes about 1 minute on average

### OS:

macOS

### PostgreSQL Version:

15.5

### DuckDB Version:

0.10.2

### DuckDB Client:

Python

### Full Name:

Mauricio Novelo

### Affiliation:

CampusESP

- [X] I agree

- [X] I agree

Mytherin:
Fixes #221 

This adds the `SCHEMA` parameter to `ATTACH`, which (1) limits the catalog entries that are loaded to only those in that schema, and (2) sets that schema as the default schema (instead of `public`). For example:

```sql
ATTACH 'dbname=postgresscanner' AS s (TYPE POSTGRES, SCHEMA 'some_schema');
SELECT * FROM s.some_table; -- reads from s.some_schema.some_table;
```

When there are many schemas in a database, this can speed up the attach as only a single schema's catalog entries are fetched.

### DIFF
--- a/src/include/storage/postgres_catalog.hpp
+++ b/src/include/storage/postgres_catalog.hpp
@@ -20,7 +20,7 @@ class PostgresSchemaEntry;
 
 class PostgresCatalog : public Catalog {
 public:
-	explicit PostgresCatalog(AttachedDatabase &db_p, const string &path, AccessMode access_mode);
+	explicit PostgresCatalog(AttachedDatabase &db_p, const string &path, AccessMode access_mode, string schema_to_load);
 	~PostgresCatalog();
 
 	string path;
@@ -93,6 +93,7 @@ private:
 	PostgresVersion version;
 	PostgresSchemaSet schemas;
 	PostgresConnectionPool connection_pool;
+	string default_schema;
 };
 
 } // namespace duckdb
--- a/src/include/storage/postgres_index_set.hpp
+++ b/src/include/storage/postgres_index_set.hpp
@@ -20,7 +20,7 @@ public:
 	PostgresIndexSet(PostgresSchemaEntry &schema, unique_ptr<PostgresResultSlice> index_result = nullptr);
 
 public:
-	static string GetInitializeQuery();
+	static string GetInitializeQuery(const string &schema = string());
 
 	optional_ptr<CatalogEntry> CreateIndex(ClientContext &context, CreateIndexInfo &info, TableCatalogEntry &table);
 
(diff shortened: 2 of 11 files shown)


mnovelo:
Really appreciate this @Mytherin! Would this be included in the next minor version upgrade of DuckDB?

PR #275, opened 2024-12-13
apply patches

Maxxen:

PR #281, opened 2025-01-08
Simplify build, move to new extension template, fixup CI, apply patches

Maxxen:
This PR introduces a number of changes to the build.

- Remove the pgconfigure bash-script. We now uses CMakes native capabilities to download/verify/unpack/configure the required PostgreSQL source code at CMake configure-time. The dependency tracking here is maybe not ideal but it at least doesn't re-run every time you do make-debug. And it works cross-platform. 

- Use extension-ci-tools to setup the makefile and extension distribution CI. 

- Don't use manylinux for the integration test CI, instead we now just use latest ubuntu and a pre-made action to setup postgresql. The pre-made action also supports windows/macOS which could be interesting to run in a matrix in the future. We also use ccache in this run now as well. 

- Updated DuckDB and applies upstream patches.

-----

Note that the integration test CI has not been run in a while so it seems like some failures that have crept in unnoticed, such as newly added table filter types which we don't support in the extension yet. These should be fixed in a separate PR.

Supersedes #275 #276


carlopi:
Looks to make a lots of sense to me!

Mytherin:
Thanks!

PR #297, opened 2025-02-07
Fix for dynamic filters generated through top-n

Mytherin:


src/postgres_filter_pushdown.cpp
```diff
@@ -98,6 +104,8 @@ string PostgresFilterPushdown::TransformFilter(string &column_name, TableFilter
 		}
 		return column_name + " IN (" + in_list + ")";
 	}
+	case TableFilterType::DYNAMIC_FILTER:
+		return string();
```
JelteF, line 108: In pg_duckdb we used a slightly different approach: We do this for any unsupported filter type that's inside an optional filter.

https://github.com/duckdb/pg_duckdb/blob/a9faa3cdb9d36881f793fef4fb519ebc675e4a6d/src/scan/postgres_scan.cpp#L66-L69

PR #313, opened 2025-04-03
Fix #233 Add limit pushdown to attached queries

### LINKED ISSUE #233: LIMIT is not pushed down
### What happens?

If I do `SELECT * FROM pg.very_large_table LIMIT 1;` the query issued does not include a limit and thus the entire table is copied.

### To Reproduce

```sql
D ATTACH '' AS pg (TYPE POSTGRES);
D SELECT * FROM pg.some_table; -- warm up schema cache
D SET pg_debug_show_queries = true;
D SELECT * FROM pg.schema_name.very_large_table LIMIT 1;
BEGIN TRANSACTION ISOLATION LEVEL REPEATABLE READ

        COPY (SELECT <lots of columns> FROM "schema_name"."very_large_table" ) TO STDOUT (FORMAT binary);
```

I don't see any pagination happening either.

My network usage goes through the roof as well (indicating lots of data is being copied).

### OS:

MacOS

### PostgreSQL Version:

15 (timescale)

### DuckDB Version:

v0.10.4-dev124 cf5b770ccb

### DuckDB Client:

CLI

### Full Name:

Adrian Garcia Badaracco

### Affiliation:

Pydantic

- [X] I agree

- [X] I agree

noahisaksen:
Implements limit pushdown https://github.com/duckdb/duckdb-postgres/issues/233

Gets applied on all postgres attached queries, if preferred can be applied only when an option like experimental pushdown is set, however in this implementation triggers the same as the mysql_scanner limit pushdown implementation. 

This change greatly improves performance for queries that include a `LIMIT`. 

What it does:

`select bar from attached_pg.foo limit 1;`

Previously

`COPY (SELECT "bar" FROM "attached_pg"."foo" ) TO STDOUT (FORMAT "binary");`

Now

`COPY (SELECT "bar" FROM "attached_pg"."foo"  LIMIT 1) TO STDOUT (FORMAT "binary");`

------

I am not sure how a specific test for this would look like as I am not sure there is a way to verify generated SQL. A generic limit test was added. 

------

Works for `LIMIT` and `OFFSET`. If an `ORDER_BY` is present it will not be added.


Mytherin:
Thanks for the PR!

Could you just fix the merge conflict?

As for testing, you can use the `EXPLAIN` to verify the limit is no longer part of the DuckDB plan, e.g.:

```sql
statement ok
set explain_output='optimized_only'

query II
EXPLAIN FROM s.large_tbl LIMIT 5;
----
logical_opt	<!REGEX>:.*"name": "LOGICAL_LIMIT".*
```

noahisaksen:
@Mytherin Thanks for the tip on the Query Plan test! 

Fixed the merge conflict and added a test to verify the LIMIT is not inside the DuckDB plan.

Mytherin:
Thanks!

PR #325, opened 2025-05-21
Add support some commonly used types for legacy text protocol

HuyNguyen7994:
Add support for version 8 TEXT protocol without COPY, so the extension could also be used to query from Redshift.
The reason to mark this as `legacy` is because Postgres and Redshift TEXT protocol has started to diverge:
- Postgres 17 starts supporting getting result in chunk: https://www.postgresql.org/docs/17/libpq-single-row-mode.html
- Redshift supports `client_protocol_veresion=3` parameter, which will transfer some data types in more efficient binary format: https://www.amazonaws.cn/en/new/2022/amazon-redshift-open-source-odbc-driver-binary-protocol-support-performance/

I only use Duckdb as a light-weight, local Redshift for unit-testing and verification, so making this super fast and efficient is not that useful.

### DIFF
--- a/create-postgres-tables.sh
+++ b/create-postgres-tables.sh
@@ -28,6 +28,7 @@ rm -rf /tmp/postgresscannertmp
 
 psql -d postgresscanner < test/all_pg_types.sql
 psql -d postgresscanner < test/decimals.sql
+psql -d postgresscanner < test/legacy_text_protocol.sql
 psql -d postgresscanner < test/other.sql
 
 
--- a/src/include/postgres_text_reader.hpp
+++ b/src/include/postgres_text_reader.hpp
@@ -1,0 +1,198 @@
+//===----------------------------------------------------------------------===//
+//                         DuckDB
+//
+// postgres_text_reader.hpp
+//
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "duckdb.hpp"
+#include "duckdb/common/types/interval.hpp"
+#include "libpq-fe.h"
+#include "postgres_conversion.hpp"
+#include "postgres_connection.hpp"
+#include <cstring>
+
+namespace duckdb {
+
+struct PostgresTextReader {
+	explicit PostgresTextReader(PostgresConnection &con_p) : con(con_p), current_result(nullptr), 
+                                                         current_row(0), current_field(0) {
+	}
+	
+	~PostgresTextReader() {
+		Reset();
+	}
+	
+	PostgresConnection &GetConn() {
+		return con;
+	}
+
+	bool ReadTextFrom(const string &query) {
+		Reset();
+		current_result = PQexec(con.GetConn(), query.c_str());
+		if (!current_result) {
+			throw IOException("Failed to execute query: %s", string(PQerrorMessage(con.GetConn())));
+		}
+		
+		auto status = PQresultStatus(current_result);
+		if (status != PGRES_TUPLES_OK && status != PGRES_SINGLE_TUPLE) {
+			string error_msg = string(PQresultErrorMessage(current_result));
+			Reset();
+			throw IOException("Query failed: %s", error_msg);
+		}
+		
+		rows = PQntuples(current_result);
+		fields = PQnfields(current_result);
+		current_row = 0;
+		current_field = 0;
+		
+		return rows > 0;
+	}
+
+	bool NextRow() {
+		if (!current_result || current_row >= rows) {
+			return false;
+		}
+		current_row++;
+		current_field = 0;
+		return current_row < rows;
+	}
+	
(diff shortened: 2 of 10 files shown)


src/include/postgres_text_reader.hpp
```diff
@@ -0,0 +1,198 @@
...
+			FlatVector::SetNull(out_vec, output_offset, true);
+			return;
+		}
+		
+		char *value = PQgetvalue(current_result, current_row, current_field);
+		int len = PQgetlength(current_result, current_row, current_field);
+		
+		switch (type.id()) {
+		case LogicalTypeId::SMALLINT:
+			FlatVector::GetData<int16_t>(out_vec)[output_offset] = ReadInteger<int16_t>(value, len);
```
Mytherin, line 139: Given that we are just doing straight-forward casting from varchar to the output type, maybe it makes sense to just use DuckDB's cast by always reading into a vector of type `VARCHAR` and casting to the output type?

test/sql/scanner/legacy_text_protocol.test
```diff
@@ -0,0 +1,25 @@
...
+require-env POSTGRES_TEST_DATABASE_AVAILABLE
+
+statement ok
+CALL postgres_attach('dbname=postgresscanner');
+
+statement ok
+SET pg_use_legacy_text_protocol=true;
+
+query I
+select c1 from legacy_text_protocol limit 1;
```
Mytherin, line 16: Why do we need a separate table for this? Can't we query the existing tables?

create-postgres-tables.sh
```diff
@@ -28,6 +28,7 @@ rm -rf /tmp/postgresscannertmp
 
 psql -d postgresscanner < test/all_pg_types.sql
 psql -d postgresscanner < test/decimals.sql
+psql -d postgresscanner < test/legacy_text_protocol.sql
```
Mytherin, line 31: I don't really see the need for this extra file

test/sql/scanner/legacy_text_protocol.test
```diff
@@ -0,0 +1,25 @@
...
+----
+1
+
+query II
+select * from legacy_text_protocol;
```
Mytherin, line 21: The tests here are rather lackluster - can we add more?

Ideally we would run all existing tests with `SET pg_use_legacy_text_protocol = true`. Perhaps we can at least do that once to verify they all pass by setting the default to true?

Mytherin (commented):
Thanks for the PR! Looks good. Supporting this sounds like a good idea. Some comments below:

src/include/postgres_text_reader.hpp
HuyNguyen7994: Can you point me where to find those methods? I found for `Date` and `Timestamp`, but not for `Text` or other primitive types

src/postgres_scanner.cpp
```diff
@@ -248,6 +254,18 @@ static void PostgresInitInternal(ClientContext &context, const PostgresBindData
...
 		}
 		filter += filter_string;
+	};
+	lstate.exec = false;
+	lstate.done = false;
+	Value use_legacy_text_protocol;
+	if (context.TryGetCurrentSetting("pg_use_legacy_text_protocol", use_legacy_text_protocol)) {
+		if (BooleanValue::Get(use_legacy_text_protocol)) {
+			lstate.sql = StringUtil::Format(
+				R"(SELECT %s FROM %s.%s %s%s;)",
```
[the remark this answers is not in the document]
HuyNguyen7994: Which layer would it best go? My C++ level is novice so there's not much design effort going into this (yet).

test/sql/scanner/legacy_text_protocol.test
HuyNguyen7994: This is just a placeholder. I also want to rerun the existing tests but with this setting on, but don't know how to accomplish this without copying the whole test set. I read the test for binary copy and it didn't attempt to rerun all cases either:
https://github.com/duckdb/duckdb-postgres/blob/main/test/sql/misc/postgres_binary.test

create-postgres-tables.sh
HuyNguyen7994: agree, I will test with existing tables then.

test/sql/scanner/legacy_text_protocol.test
Mytherin: We can just add a test for all types to the main test suite - but I would set `pg_use_legacy_text_protocol` to true by default and then manually run the tests as a verification pass.

HuyNguyen7994: yeah, I just did that on my branch. Will attach the CI result there once I managed to cover all the cases.

PR #343, opened 2025-07-16
Fix various memory leaks

JelteF:
In a bunch of cases PQclear was not being called on a (possibly)
initialized `PGresult *`. This fixes those cases. These are all fairly
small memory leaks, but they could add up over time if multiple queries
to Postgres are being done.


Mytherin:
Thanks!

Mytherin:
After looking at this a bit more closely this introduces a number of use-after-free bugs with reading error messages from results after they are destroyed - I've switched to using RAII to destroy the results in https://github.com/duckdb/duckdb-postgres/pull/346 instead

PR #375, opened 2025-10-01
Add postgres secret store

aclowes:
The goal is to store the secrets in postgres so they can be retrieved remotely and automatically when attaching a lake. For example, the initial configuration is:
```
ATTACH 'postgres:postgresql://user:password@host/lake' AS metadata;
ATTACH 'ducklake:postgres:postgresql://user:password@host/lake' AS lake (DATA_PATH 's3://some-bucket');
CREATE SECRET IN postgres_metadata (
	TYPE S3,
	KEY_ID 'test_key_id',
	SECRET 'test_secret_value',
	REGION 'us-east-1'
);
```
After which, duckdb can be run in the JDBC driver for BI tools with a connection string like this, and it will be able to connect to postgres and S3 without further configuration:
```
jdbc:duckdb:ducklake:postgres:postgresql://user:password@host/lake
```
Feedback appreciated!

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -15,6 +15,7 @@ add_library(
   postgres_filter_pushdown.cpp
   postgres_query.cpp
   postgres_scanner.cpp
+  postgres_secret_storage.cpp
   postgres_storage.cpp
   postgres_text_reader.cpp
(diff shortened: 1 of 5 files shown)


staticlibs:
Closing as superseded by #461

PR #386, opened 2025-11-21
Feat/rds auth

### LINKED ISSUE #360: Feature Request: Automatically get password for AWS RDS IAM based authentication
I have an RDS Aurora PostgreSQL cluster with IAM-based authentication enabled. To authenticate, I need to call the [GenerateRDSAuthToken](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/example_rds_GenerateRDSAuthToken_section.html) API to generate a temporary token to use as the password. Here's an example CLI command.

```
aws rds generate-db-auth-token --hostname some-postgres-host --port 5432 --region us-west-2 --username some-username
```

The tokens only last 15 minutes. It would be really helpful if there was a way to configure a `postgres` secret to get the password through the above API call or by invoke a CLI command.

pquadri:
Fixes https://github.com/duckdb/duckdb-postgres/issues/360.

Quite dirty, but it seemed the least intrusive solution, let me know what you think about this. 
It can potentially be just an experimental addition then be changed to use the AWS SDK? (i didn't want to add that as a dependency, having a soft requirement of "you need aws-cli if you want to use this" seemed better)

I'm working on having an RDS with IAM to test this properly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,8 @@ endif()
 include_directories(
     include
     database-connector/src/include
-    ${OPENSSL_INCLUDE_DIR})
+    ${OPENSSL_INCLUDE_DIR}
+    ${PostgreSQL_INCLUDE_DIRS})
 
 if(WIN32)
   include_directories(
--- a/README.md
+++ b/README.md
@@ -52,6 +52,53 @@ D SELECT * FROM postgres_db.uuids;
 
 For more information on how to use the connector, refer to the [Postgres documentation on the website](https://duckdb.org/docs/extensions/postgres).
 
+### AWS RDS IAM Authentication
+
+The extension supports AWS RDS IAM-based authentication, which allows you to connect to RDS PostgreSQL instances using IAM database authentication instead of static passwords. This feature automatically generates temporary authentication tokens using the AWS CLI.
+
+#### Requirements
+
+- AWS CLI installed and configured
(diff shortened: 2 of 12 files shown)


staticlibs:
Hi, thanks for the PR!

> It can potentially be just an experimental addition then be changed to use the AWS SDK?

M, I cannot answer this off the top of my head, there are obviously good and bad parts of bringing in AWS SDK.

Meanwhile, to pass the CI format check

1. install `black==25.1.0`, `clang_format==11.0.1` and `cmake-format==0.6.13` from Python pip
2. run `make format-fix`

Or, alternatively, just apply the diff from the CI job output.

pquadri:
> Hi, thanks for the PR!
> 
> > It can potentially be just an experimental addition then be changed to use the AWS SDK?
> 
> M, I cannot answer this off the top of my head, there are obviously good and bad parts of bringing in AWS SDK.
> 
> Meanwhile, to pass the CI format check
> 
> 1. install `black==25.1.0`, `clang_format==11.0.1` and `cmake-format==0.6.13` from Python pip
> 2. run `make format-fix`
> 
> Or, alternatively, just apply the diff from the CI job output.

sorry missed that 😅 
i also have fixed feelings about the extra dependency. I think there is potentially another option to depend on the aws extension instead, but that also doesn't fully convince me

staticlibs:
Closing as superseded by #457.

PR #396, opened 2026-01-04
Fix #395 - LIMIT Pushdown -> force single task execution

### LINKED ISSUE #395: LIMIT 1 returning multiple rows within open transaction
### What happens?

I'm finding that tables of a sufficient size are returning multiple rows even after specifying `LIMIT 1`. For me, it only seems to happen if I've tried to create an index within the same transaction. After committing the transaction, the expected behaviour returns.

This seems similar to #380 , #381 but I'm not sure so I opened a new issue.

Notes:
- On the test table, it did not show up with tables of sizes 5000, 500000.
- On my "real data", it shows up on a table with ~238000 rows
- On the test table, a size of 1000000 seems to return two rows, and 3000000 returns 6 rows.
- I was not able to reproduce this on a standard duckdb table
- I was not able to reproduce this directly against the postgres database (e.g. via `psql`). 
- I'm running DuckDB via `nix`, but 

### To Reproduce

Start a postgres server. I don't think this is specific to the podman or docker image because the same happens with postgres via a nix flake. I've included a podman command for ease of replication.

```
podman run -it -e POSTGRES_HOST_AUTH_METHOD=trust -e POSTGRES_USER=duck -e POSTGRES_DB=duck --network=host -p 5432:5432 docker.io/postgres:18.
```

Attach to the database in duckdb

```
INSTALL postgres; LOAD postgres;
ATTACH 'user=duck dbname=duck host=127.0.0.1 port=5432' AS pg (TYPE postgres);
```

Create a table on pg

```
CREATE TABLE pg.test_table AS
WITH data as (
    SELECT
        unnest(generate_series(1,3000000)) as id,
        floor(random() * 3)::int as grp
)
select
    data.id,
    CASE
        WHEN data.grp = 0 then 'Group A'
        WHEN data.grp = 1 then 'Group B'
        WHEN data.grp = 2 then 'Group C'
    END as grp
from data;
;
```

Observe that without an index, limit 1 works

```
SELECT * FROM pg.test_table WHERE grp = 'Group A' limit 1;
```
```
┌───────┬─────────┐
│  id   │   grp   │
│ int64 │ varchar │
├───────┼─────────┤
│   0   │ Group A │
└───────┴─────────┘
```

Create an index within a transaction and query limit 1 within the transaction

```
BEGIN;
CREATE INDEX test_idx ON pg.test_table (grp);
SELECT * FROM pg.test_table WHERE grp = 'Group A' limit 1;
```
```
┌────────┬─────────┐
│   id   │   grp   │
│ int64  │ varchar │
├────────┼─────────┤
│      0 │ Group A │
│ 185000 │ Group A │
│ 370001 │ Group A │
│ 555000 │ Group A │
│ 740010 │ Group A │
│ 925000 │ Group A │
└────────┴─────────┘
```

I don't think an indexed is being used, but I can't seem to recreate this without the index (`EXPLAIN ANALYZE` was run inside of the transaction)
```
┌─────────────────────────────────────┐
│┌───────────────────────────────────┐│
││    Query Profiling Information    ││
│└───────────────────────────────────┘│
└─────────────────────────────────────┘
EXPLAIN ANALYZE SELECT * FROM pg.test_table WHERE grp = 'Group A' limit 1;
┌────────────────────────────────────────────────┐
│┌──────────────────────────────────────────────┐│
││              Total Time: 0.0015s             ││
│└──────────────────────────────────────────────┘│
└────────────────────────────────────────────────┘
┌───────────────────────────┐
│           QUERY           │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│      EXPLAIN_ANALYZE      │
│    ────────────────────   │
│           0 rows          │
│          (0.00s)          │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│         PROJECTION        │
│    ────────────────────   │
│             id            │
│            grp            │
│                           │
│           6 rows          │
│          (0.00s)          │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│         TABLE_SCAN        │
│    ────────────────────   │
│     Table: test_table     │
│                           │
│        Projections:       │
│            grp            │
│             id            │
│                           │
│          Filters:         │
│       grp='Group A'       │
│                           │
│           6 rows          │
│          (0.00s)          │
└───────────────────────────┘
```

Aborting the transaction yields the correct results

```
rollback;
SELECT * FROM pg.test_table WHERE grp = 'Group A' limit 1;
┌───────┬─────────┐
│  id   │   grp   │
│ int64 │ varchar │
├───────┼─────────┤
│   0   │ Group A │
└───────┴─────────┘
```

Interestingly, trying to recreate the index outside of the aborted transaction yields an error, but not sure if related. I'm not able to drop the index either, 

```
D CREATE INDEX test_idx ON pg.test_table (grp);
Binder Error:
Index with name "test_idx" already exists in schema "public
D DROP INDEX test_idx;
Catalog Error:
Index with name test_idx does not exist!
Did you mean "pg.test_idx"?
D DROP INDEX pg.test_idx;
Invalid Error:
Failed to execute query "BEGIN TRANSACTION ISOLATION LEVEL REPEATABLE READ;
DROP Index "public"."test_idx"": ERROR:  index "test_idx" does not exist
```

If you repeat all the steps above on a fresh db, but commit instead of rolling back, you get the expected behaviour

```
D INSTALL postgres; LOAD postgres;
D ATTACH 'user=duck dbname=duck host=127.0.0.1 port=5432' AS pg (TYPE postgres);
D CREATE TABLE pg.test_table AS
  WITH data as (
      SELECT
          unnest(generate_series(3000000)) as id,
          floor(random() * 3)::int as grp
  )
  select
      data.id,
      CASE
          WHEN data.grp = 0 then 'Group A'
          WHEN data.grp = 1 then 'Group B'
          WHEN data.grp = 2 then 'Group C'
      END as grp
  from data;
D ;
D SELECT * FROM pg.test_table WHERE grp = 'Group A' limit 1;
┌───────┬─────────┐
│  id   │   grp   │
│ int64 │ varchar │
├───────┼─────────┤
│   0   │ Group A │
└───────┴─────────┘
D BEGIN;
D CREATE INDEX test_idx ON pg.test_table (grp);
D SELECT * FROM pg.test_table WHERE grp = 'Group A' limit 1;
┌────────┬─────────┐
│   id   │   grp   │
│ int64  │ varchar │
├────────┼─────────┤
│      0 │ Group A │
│ 185000 │ Group A │
│ 370002 │ Group A │
│ 555002 │ Group A │
│ 740000 │ Group A │
│ 925000 │ Group A │
└────────┴─────────┘
D commit;
D SELECT * FROM pg.test_table WHERE grp = 'Group A' limit 1;
┌───────┬─────────┐
│  id   │   grp   │
│ int64 │ varchar │
├───────┼─────────┤
│   0   │ Group A │
└───────┴─────────┘
D EXPLAIN ANALYZE select * from pg.test_table WHERE grp = 'Group A' limit 1;
┌─────────────────────────────────────┐
│┌───────────────────────────────────┐│
││    Query Profiling Information    ││
│└───────────────────────────────────┘│
└─────────────────────────────────────┘
EXPLAIN ANALYZE select * from pg.test_table WHERE grp = 'Group A' limit 1;
┌────────────────────────────────────────────────┐
│┌──────────────────────────────────────────────┐│
││              Total Time: 0.0028s             ││
│└──────────────────────────────────────────────┘│
└────────────────────────────────────────────────┘
┌───────────────────────────┐
│           QUERY           │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│      EXPLAIN_ANALYZE      │
│    ────────────────────   │
│           0 rows          │
│          (0.00s)          │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│         PROJECTION        │
│    ────────────────────   │
│             id            │
│            grp            │
│                           │
│           1 row           │
│          (0.00s)          │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│      STREAMING_LIMIT      │
│    ────────────────────   │
│           1 row           │
│          (0.00s)          │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│         TABLE_SCAN        │
│    ────────────────────   │
│     Table: test_table     │
│                           │
│        Projections:       │
│            grp            │
│             id            │
│                           │
│          Filters:         │
│       grp='Group A'       │
│                           │
│         4,096 rows        │
│          (0.00s)          │
└───────────────────────────┘
```

I am also able to reproduce this on the latest main. To do that, I fired up a python docker container and installed via pip.

```
podman run --network=host -it docker.io/python:3.14-bookworm /bin/bash
pip install duckdb --upgrade --pre
```

and then running this inside the repl in the container:
```
import duckdb
con = duckdb.connect(":memory:")
con.sql("SELECT version()").show()
con.sql("INSTALL postgres; LOAD postgres; ATTACH 'user=duck dbname=duck host=127.0.0.1 port=5432' AS pg (TYPE postgres);")

con.sql("""
CREATE TABLE pg.test_table AS
WITH data as (
    SELECT
        unnest(generate_series(1,3000000)) as id,
        floor(random() * 3)::int as grp
)
select
    data.id,
    CASE
        WHEN data.grp = 0 then 'Group A'
        WHEN data.grp = 1 then 'Group B'
        WHEN data.grp = 2 then 'Group C'
    END as grp
from data;
;
""")

con.sql("SELECT * FROM pg.test_table WHERE grp = 'Group A' limit 1;").show()
con.begin()
con.sql("CREATE INDEX test_idx ON pg.test_table (grp);")
con.sql("SELECT * FROM pg.test_table WHERE grp = 'Group A' limit 1;").show()
con.rollback()
con.sql("SELECT * FROM pg.test_table WHERE grp = 'Group A' limit 1;").show()
```

### OS:

Fedora Silverblue 41

### PostgreSQL Version:

18.1

### DuckDB Version:

1.4.1

### DuckDB Client:

CLI

### Full Name:

Dillon Chan

### Affiliation:

N/A

- [x] I agree

- [x] I agree

noahisaksen:
To fix `LIMIT` returning more than the expected rows, this PR guarantees single-task/thread execution when LIMIT Pushdown occurs. 

Relevant issues:  https://github.com/duckdb/duckdb-postgres/issues/380 https://github.com/duckdb/duckdb-postgres/issues/395

BACKGROUND: 

LIMIT Pushdown introduced big optimization but also an issue where `LIMIT N` would return more than `N` results, due to some thread and tasking. 

Initially fixed in https://github.com/duckdb/duckdb-postgres/pull/330 , where we only apply LIMIT Pushdown, if `max_threads == 1`. This did not cover all cases. 

As suggested by @Mytherin already on the PR above, it may be better to set threads and tasks to 1, when LIMIT optimization can be applied. This PR does that. 

---

Before fixes applied, `Test 4` inside `test/sql/storage/limit_parallel_bug.test` was failing by returning more than 1 row. To recreate https://github.com/duckdb/duckdb-postgres/issues/395

---

This PR removes the check made in https://github.com/duckdb/duckdb-postgres/pull/330 in favor of setting the amount of threads and tasks.

### DIFF
--- a/src/storage/postgres_optimizer.cpp
+++ b/src/storage/postgres_optimizer.cpp
@@ -53,11 +53,6 @@ static void OptimizePostgresScanLimitPushdown(unique_ptr<LogicalOperator> &op) {
 		}
 
 		auto &bind_data = get.bind_data->Cast<PostgresBindData>();
-		if (bind_data.max_threads != 1 || !bind_data.can_use_main_thread) {
-			// cannot push down limit/offset if we are not using the main thread
-			OptimizePostgresScanLimitPushdown(op->children[0]);
-			return;
-		}
 
 		string generated_limit_clause = "";
 		if (limit.limit_val.Type() != LimitNodeType::UNSET) {
--- a/src/storage/postgres_optimizer.cpp
+++ b/src/storage/postgres_optimizer.cpp
@@ -69,6 +69,11 @@ static void OptimizePostgresScanLimitPushdown(unique_ptr<LogicalOperator> &op) {
 
 		if (!generated_limit_clause.empty()) {
 			bind_data.limit = generated_limit_clause;
+			// When LIMIT is pushed down to Postgres, we must ensure single-task execution
+			// to avoid each task (whether parallel or sequential) applying the LIMIT independently.
+			// Setting pages_approx = 0 disables CTID-based task splitting, ensuring a single query.
+			bind_data.pages_approx = 0;
+			bind_data.max_threads = 1;
 
 			op = std::move(op->children[0]);
 			return;
--- a/test/sql/storage/limit.test
+++ b/test/sql/storage/limit.test
@@ -65,9 +65,10 @@ FROM s.large_tbl LIMIT 5 OFFSET 5
 statement ok
 set explain_output='optimized_only'
 
-# limit is still in plan as we were not able to push down due to parallel execution
+# limit is pushed down to Postgres even with pg_pages_per_task=1
+# (we now force single-task execution when LIMIT is pushed down)
 
 query II
 EXPLAIN FROM s.large_tbl LIMIT 5;
 ----
-logical_opt	<REGEX>:.*LIMIT.*
+logical_opt	<!REGEX>:.*LIMIT.*
--- a/test/sql/storage/limit_parallel_bug.test
+++ b/test/sql/storage/limit_parallel_bug.test
@@ -62,6 +62,5 @@ SELECT count(*) FROM (SELECT * FROM s.limit_test_large WHERE grp = 'Group A' LIM
 statement ok
 ROLLBACK
 
-# Cleanup
 statement ok
 DROP TABLE IF EXISTS s.limit_test_large
--- a/test/sql/storage/limit_parallel_bug.test
+++ b/test/sql/storage/limit_parallel_bug.test
@@ -1,0 +1,67 @@
+# name: test/sql/storage/limit_parallel_bug.test
+# description: Test LIMIT bug with parallelism (issues #380, #395)
+# group: [storage]
+
+require postgres_scanner
+
+require-env POSTGRES_TEST_DATABASE_AVAILABLE
+
+statement ok
+ATTACH 'host=127.0.0.1 port=5432 dbname=postgresscanner user=postgres' AS s (TYPE POSTGRES)
+
+# Create a large table to enable parallelism
+statement ok
+CREATE OR REPLACE TABLE s.limit_test_large AS
+SELECT unnest(generate_series(1, 1000000)) as id,
+       CASE floor(random() * 3)::int
+           WHEN 0 THEN 'Group A'
+           WHEN 1 THEN 'Group B'
+           ELSE 'Group C'
+       END as grp
+
+# Force parallelism with small pages_per_task
+statement ok
+SET pg_pages_per_task=1
+
+# Test 1: Basic LIMIT 1 should return exactly 1 row
+query I
+SELECT count(*) FROM (SELECT * FROM s.limit_test_large LIMIT 1)
+----
+1
+
+# Test 2: LIMIT 1 with WHERE clause
+query I
+SELECT count(*) FROM (SELECT * FROM s.limit_test_large WHERE grp = 'Group A' LIMIT 1)
+----
+1
+
+# Test 3: Within transaction
+statement ok
+BEGIN
+
+query I
+SELECT count(*) FROM (SELECT * FROM s.limit_test_large LIMIT 1)
+----
+1
+
+statement ok
+ROLLBACK
+
+# Test 4: Transaction with index creation (issue #395 scenario)
+statement ok
+BEGIN
+
+statement ok
+CREATE INDEX IF NOT EXISTS limit_test_idx ON s.limit_test_large (grp)
+
+query I
+SELECT count(*) FROM (SELECT * FROM s.limit_test_large WHERE grp = 'Group A' LIMIT 1)
+----
+1
+
+statement ok
+ROLLBACK
+
+# Cleanup
+statement ok
+DROP TABLE IF EXISTS s.limit_test_large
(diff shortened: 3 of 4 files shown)


staticlibs (approved):
Thanks for the PR! Looks good to me, sorry for the delay!

Issue #366, opened 2025-08-29
postgres_scanner at commit 03eaed75… (DuckDB 1.1.3) lacks extension_config.cmake and static target — cannot bundle without DONT_LINK


nadirgozcu:
### Context
I’m building a custom libduckdb_bundle.a for DuckDB 1.1.3, bundling with extensions (duckpgq and postgres_scanner / duckdb-postgres) to produce a fully static binary (FROM scratch). The target commit for Postgres is 03eaed75f0ec5500609b7a97aa05468493b229d1. According to:
https://github.com/duckdb/duckdb/blob/v1.1.3/.github/config/out_of_tree_extensions.cmake#L96

### Issue
At this commit:

There is no extension_config.cmake in duckdb-postgres, so the DuckDB can’t discover a static target out-of-the-box.

I manually created the `extension_config.cmake` file as in main branch:
```
duckdb_extension_load(postgres_scanner
    SOURCE_DIR ${CMAKE_CURRENT_LIST_DIR}
    DONT_LINK
    LOAD_TESTS
)
```

The extension’s CMake path uses DONT_LINK, which forces a loadable (.duckdb_extension) build only.

If I remove DONT_LINK, the superbuild expects a static target named postgres_scanner_extension, but link fails with:

cannot find -lpostgres_scanner_extension

If I keep DONT_LINK, the extension builds as loadable, but it is not included in libduckdb_bundle.a. Verifying the bundle:

ar -t libduckdb_bundle.a | grep -i postgres || true
(no output)

So I cannot statically link it.

### What I need
One of the following for the 03eaed75… line (DuckDB 1.1.3 target):

An extension_config.cmake plus a static library target (e.g., postgres_scanner_extension) so the DuckDB can link it into libduckdb_bundle.a, or

Official confirmation that static linking is not supported for this commit (an also in main version?), with guidance on a tag/commit that does support static, or the minimal patch needed to add a static target.

### Why this matters
I need a fully static final binary (no dlopen at runtime) to run on FROM scratch. Loadable-only (DONT_LINK) forces a dynamic base image (or shipping many runtime libs), which we’re trying to avoid.

### Repro (simplified)

Core: DuckDB v1.1.3

Extensions: duckdb-postgres (03eaed75…)

Build:

```
make bundle-library DUCKDB_PLATFORM=any \
  EXTENSION_CONFIGS="/path/to/duckdb-postgres/extension_config.cmake"
```

[...]

PR #402, opened 2026-02-03
Add static build & add `-Bsymbolic` for loadable extension

YuweiXiao:
Changes:

- Add static build 
- Add `-Bsymbolic` for loadable extension build

When `postgres_scanner` is loaded inside a PostgreSQL backend (e.g., via pg_duckdb), the connection to postgres may fail with:
```
libpq is incorrectly linked to backend functions
```

The cause: postgres_scanner statically compiles libpq with `-DFRONTEND=1`, so function in libpq, i.e., `pg_link_canary_is_frontend()`, would validate if it is running under FRONTEND during connection startup. However, on ELF systems, the dynamic linker resolves the function symbol to the PostgreSQL backend's version, which is not FRONTEND, leading to connection failure.

By adding `-Bsymbolic`, we enforce postgres_scanner to bind internal symbol references within the extension.


YuweiXiao:
this PR shall solve #366

staticlibs:
Hi, thanks for the PR! I'll close and reopen it to run the CI.

On the contents of the PR - I wonder whether we need to pass the `C_STANDARD 99` to the static extension target too?

Issue #415, opened 2026-03-19
GEOMETRY column not recognized when using postgres_query()


peterschaer:
### What happens?

I have a table in PostgreSQL with a GEOMETRY column. Since DuckDB 1.5, this GEOMETRY column is recognized during import:

```
select objectid, shape from nupla_uept limit 5;

┌──────────┬─────────────────────────────────┐
│ objectid │              shape              │
│  int32   │            geometry             │
├──────────┼─────────────────────────────────┤
│   116927 │ POINT (2604833.296 1208390.399) │
│   116928 │ POINT (2605120.181 1208221.813) │
│   116929 │ POINT (2604777.003 1207929.125) │
│    98169 │ POINT (2610375.849 1189134.644) │
│    98170 │ POINT (2610409.164 1189117.758) │
└──────────┴─────────────────────────────────┘
```
However, if I make the query with `postgres_query()`, the column is not recognized:

```
select * from postgres_query('vek1', 'SELECT objectid, shape from geodb.nupla_uept') LIMIT 5;

┌──────────┬────────────────────────────────────────────────────┐
│ objectid │                       shape                        │
│  int32   │                      varchar                       │
├──────────┼────────────────────────────────────────────────────┤
│   116927 │ 010100002008080000F853E3A590DF43412FDD246646703241 │
│   116928 │ 0101000020080800000C022B1720E043419CC420D09D6F3241 │
│   116929 │ 010100002008080000D34D628074DF434100000020796E3241 │
│    98169 │ 0101000020080800003108ACEC63EA43411B2FDDA40E253241 │
│    98170 │ 010100002008080000B6F3FD9474EA4341BA490CC2FD243241 │
└──────────┴────────────────────────────────────────────────────┘
```
```
It does not work with this generic statement either:

select * from postgres_query('vek1', 'select st_geometryfromtext(''POINT(2600000 1200000)'', 2056);');

┌────────────────────────────────────────────────────┐
│                st_geometryfromtext                 │
│                      varchar                       │
├────────────────────────────────────────────────────┤
│ 0101000020080800000000000020D6434100000000804F3241 │
```

Is there a reason for this different behavior?

### To Reproduce

1. Attach a PostgreSQL Database with the PostGIS Extension installed
2. Execute `select * from postgres_query('vek1', 'select st_geometryfromtext(''POINT(2600000 1200000)'', 2056);');`

### OS:

Windows 11

### PostgreSQL Version:

15.14

### DuckDB Version:

[...]

PR #418, opened 2026-03-20
feat: propagate GEOMETRY CRS through PostGIS in both directions

jatorre:
## Summary

Make DuckDB 1.5's GEOMETRY CRS metadata (`GEOMETRY('EPSG:N')`) survive a full round trip through PostgreSQL/PostGIS via `postgres_scanner`.

### Write side (DuckDB → PostGIS)

`PostgresBinaryWriter::WriteGeometry` detects CRS on the `LogicalType` and writes EWKB instead of plain WKB so PostGIS receives the SRID:

```
WKB:  [byte_order:1] [type:4]                    [payload...]
EWKB: [byte_order:1] [type|0x20000000:4] [srid:4] [payload...]
```

Falls back to plain WKB when no CRS is set, preserving previous behaviour. Uses `GeoType::HasCRS` / `GeoType::GetCRS` (the earlier draft mistakenly referenced `LogicalType::GetCRS`, which does not exist on the public DuckDB 1.5 API).

### Read side (PostGIS → DuckDB)

`PostgresUtils::TypeToLogicalType` decodes the SRID from PostGIS's column-level typmod for `geometry(TYPE, SRID)` columns and returns `LogicalType::GEOMETRY("EPSG:N")` instead of plain `GEOMETRY()`. The typmod layout (PostGIS source):

| bits | meaning |
|------|---------|
| 0 | has-M |
| 1 | has-Z |
| 2–7 | geometry type code (POINT=1, LINESTRING=2, …) |
| 8–29 | SRID |

Untyped `geometry` columns carry typmod -1 and fall back to plain `GEOMETRY` (no CRS), matching the previous behaviour.

### Before / after

| | Before | After |
|---|---|---|
| Write `GEOMETRY('EPSG:4326')` | PostGIS `geometry(SRID=0)` | PostGIS `geometry(SRID=4326)` |
| Read `geometry(POINT, 4326)` | DuckDB `GEOMETRY` (no CRS) | DuckDB `GEOMETRY('EPSG:4326')` |

### Test plan

- [x] Unit tests build (`make release`).
- [x] `clang-format` clean (was a CI failure on the previous draft).
- [x] End-to-end SRID survival via a DuckDB ATTACH against live PostGIS:
  - `npm run test:adbc-srid -- postgresql --srid=3857` passes — `remote_srid=3857`, `duckdb_crs=EPSG:3857`, `duckdb_type=GEOMETRY('EPSG:3857')`.
  - Multi-shape (point, line, polygon, NULL) byte-for-byte WKB equality round trip passes.

### DIFF
--- a/src/include/postgres_binary_writer.hpp
+++ b/src/include/postgres_binary_writer.hpp
@@ -10,6 +10,7 @@
 
 #include "duckdb.hpp"
 #include "duckdb/common/types/interval.hpp"
+#include "duckdb/common/types/geometry_crs.hpp"
 #include "duckdb/common/serializer/memory_stream.hpp"
 #include "postgres_conversion.hpp"
 
@@ -207,6 +208,44 @@ public:
 		stream.WriteData(const_data_ptr_cast(str_data), str_size);
 	}
 
+	//! Write GEOMETRY to PostGIS. If the type carries CRS metadata, writes EWKB
+	//! (WKB with SRID) so PostGIS receives the correct SRID.
+	void WriteGeometry(string_t value, const LogicalType &type) {
+		if (!GeoType::HasCRS(type)) {
+			WriteRawBlob(value);
+			return;
+		}
+		auto &crs = GeoType::GetCRS(type);
+
+		// Extract SRID from CRS identifier (e.g., "EPSG:4326" → 4326)
+		auto &id = crs.GetIdentifier();
+		auto colon = id.find(':');
+		if (colon == string::npos) {
+			WriteRawBlob(value);
+			return;
+		}
+		int32_t srid;
+		try {
+			srid = std::stoi(id.substr(colon + 1));
+		} catch (...) {
+			WriteRawBlob(value);
+			return;
+		}
+
+		// Write EWKB: WKB with SRID flag set on the type field + 4-byte SRID inserted.
+		// [byte_order:1] [type|0x20000000:4] [srid:4] [coordinates...]
+		auto wkb_size = value.GetSize();
+		auto wkb_data = const_data_ptr_cast(value.GetData());
+		WriteRawInteger<int32_t>(NumericCast<int32_t>(wkb_size + 4));
+		stream.WriteData(wkb_data, 1); // byte order
+		uint32_t wkb_type;
+		memcpy(&wkb_type, wkb_data + 1, 4);
+		wkb_type |= 0x20000000;
+		stream.WriteData(const_data_ptr_cast(reinterpret_cast<const uint8_t *>(&wkb_type)), 4); // type + SRID flag
+		stream.WriteData(const_data_ptr_cast(reinterpret_cast<const uint8_t *>(&srid)), 4); // SRID (LE, matching WKB)
+		stream.WriteData(wkb_data + 5, wkb_size - 5);                                       // rest of payload
+	}
+
 	void WriteVarchar(string_t value) {
 		auto str_size = value.GetSize();
 		auto str_data = value.GetData();
@@ -354,7 +393,7 @@ public:
 		}
 		case LogicalTypeId::GEOMETRY: {
 			auto data = FlatVector::GetData<string_t>(col)[r];
-			WriteRawBlob(data);
+			WriteGeometry(data, type);
 			break;
 		}
 		case LogicalTypeId::ENUM: {
--- a/src/postgres_utils.cpp
+++ b/src/postgres_utils.cpp
@@ -147,6 +147,20 @@ LogicalType PostgresUtils::TypeToLogicalType(optional_ptr<PostgresTransaction> t
 		postgres_type.info = PostgresTypeAnnotation::JSONB;
 		return LogicalType::VARCHAR;
 	} else if (pgtypename == "geometry") {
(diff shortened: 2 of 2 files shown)


jatorre:
Good news — I checked a default PostGIS `spatial_ref_sys` table and all 6,184 EPSG entries have `srid = auth_srid` with zero mismatches:

```sql
SELECT COUNT(*) as total,
       SUM(CASE WHEN srid = auth_srid::INT THEN 1 ELSE 0 END) as matching,
       SUM(CASE WHEN srid != auth_srid::INT THEN 1 ELSE 0 END) as mismatched
FROM spatial_ref_sys WHERE auth_name = 'EPSG';

-- total: 6184, matching: 6184, mismatched: 0
```

So for the default table, EPSG code = PostGIS SRID is a safe assumption. A user could theoretically insert custom rows that break this, but that's an edge case they'd need to handle themselves anyway.

Happy to wait until you have time to look at it properly — no rush.

jatorre:
Follow-up: the read path has the symmetric issue. `postgres_scanner` reads PostGIS `geometry` columns as plain WKB (type `0x01`), not EWKB (`0x01 | 0x20000000`), so the SRID is stripped on the way in too. A column typed `geometry(Point, 3857)` in PostGIS arrives in DuckDB as `GEOMETRY` with empty `ST_CRS()`.

Two pieces needed for a full fix:
1. **Bind time** — read the PostGIS typmod from `pg_attribute` (or `geometry_columns`) and bind the column as `GEOMETRY('EPSG:NNNN')` instead of plain `GEOMETRY`.
2. **Per-value** — in `postgres_binary_reader.cpp` (`LogicalTypeId::GEOMETRY` branch), strip the EWKB SRID header before passing bytes to `Geometry::FromBinary`.

Happy to follow up with a separate PR for the read side. Keeping this PR focused on the write path since the two changes touch different code and have different edge cases (untyped `geometry` columns, mixed-SRID columns, `postgres_query()` path — see #415).

jatorre:
Pushed `640b9fa` to cover a second hole I hit while testing this end-to-end through DuckDB-mediated cross-warehouse transfers.

The previous commit (`99cd2f8`) handles the EWKB write path and typmod read decode for tables created with an explicit `geometry(geometry, SRID)` column. But for `CREATE TABLE AS` — which is how DuckDB-driven `CREATE TABLE pg.public.t AS SELECT * FROM <source>` lands geometry — the SRID was still being stripped before reaching either side, so the column landed as plain `geometry` (typmod `-1`) and the round trip lost the CRS.

There are two strip points in `src/postgres_utils.cpp`:

1. **`ToPostgresType`** is called from `AddCastToPostgresTypes` during CTAS physical planning. It used to return `LogicalType::GEOMETRY()` unconditionally for any GEOMETRY input, dropping the CRS before the projection that feeds the insert. Changed to return the input unchanged so CRS survives the cast/projection.

2. **`TypeToString`** used to return plain `"GEOMETRY"` for any GEOMETRY type, and would have been short-circuited anyway by the `HasAlias()` early return (`GEOMETRY('EPSG:N')` carries an alias of `"GEOMETRY"`). The new branch runs *before* `HasAlias`, detects CRS via `GeoType::HasCRS`, and emits `geometry(Geometry, N)` so PostGIS records the SRID in `pg_attribute.atttypmod` and the existing read-side typmod decode (already on this PR) picks it up correctly.

Verified end-to-end with a small streamability harness that chains `adbc_insert(dest, ..., (SELECT * FROM <postgres>...))` through a single DuckDB connection:

| Source → Dest | Plan | WKB | CRS at dest |
|---|---|---|---|
| postgres → oracle | streams (no materialization op) | byte-match | `EPSG:3857` ✓ |
| postgres → snowflake | streams | byte-match | `EPSG:3857` ✓ |

Both round trips were failing before this commit (CRS dropped to `EPSG:4326` at the destination because the postgres source lost it). Same plan shape, no change in elapsed time, RSS stays flat.

This is the only change beyond the previous commit; the EWKB write and typmod read decode are unchanged.

jatorre:
@Maxxen picking this back up — happy to rescope the PR to the EPSG-only handling you suggested: treat `geometry(Geometry, N)` typmod SRIDs strictly as EPSG codes both on read (typmod → GeoArrow `{"crs":"EPSG:N"}`) and write (`ToPostgresType` emitting the typmod only when the CRS is an EPSG code), and leave `OGC:CRS84`/non-EPSG authorities out of scope for now. That keeps the CTAS round-trip lossless for the overwhelmingly common case without pretending SRID == any authority code.

If that scoping works for you I'll rebase onto current main (the recent DuckDB bumps and dbconnector pushdown work) and update the PR this week. We've been running this patch in production at CARTO since April for warehouse imports/exports, so it's had real mileage.

jatorre:
Hey @Maxxen any news on this front? we are running a fork that we would love to deprecate

Maxxen:
I haven't worked on this anything myself - but if this PR works then it looks good to me. I don't seem to have merge access though, maybe @staticlibs does?

staticlibs:
@jatorre 

Thanks for the PR! I am not qualified to review Spatial stuff, so thanks to Max for looking on it. I filed #579 to make this available in 1.5.5/1.5.6 from core_nightly repo.

PR #430, opened 2026-04-05
Pool: add configuration and introspection function

staticlibs:
When a Postgres database is being attached, a connection pool is created for this database.

This PR adds the ability to configure connection pool details for both newly created pools (newly attached databases) and also for already existing pools (for databases that are already attached).

It adds the following configuration options that only apply to databases attached after the option is set:

 - `pg_pool_max_connections` (`UBIGINT`, default: `max(num_cpus, 8)`): maximum number of connections that are allowed to be cached in a connection pool for each attached Postgres database. This number can be temporary exceeded when parallel scans are used.
 - `pg_pool_wait_timeout_millis` (`UBIGINT`, default: `30000`): maximum number of milliseconds to wait when acquiring a connection from a pool where all available connections are already taken.
 - `pg_pool_enable_thread_local_cache` (`BOOLEAN`, default: `true`): whether to enable the connection caching in thread-local cache. Such connections are getting pinned to the threads and are not made available to other threads, while still taking the place in the pool.
 - `pg_pool_max_lifetime_millis` (`UBIGINT`, default: `0` - unlimited): maximum number of milliseconds the connection can be kept open. This number is checked when the connection is taken from the pool and returned to the pool. When the connection pool reaper thread is enabled ('pg_pool_enable_reaper_thread' option), then this number is checked in background periodically.
 - `pg_pool_idle_timeout_millis` (`UBIGINT`, default: `0` - unlimited): maximum number of milliseconds the connection can be kept idle in the pool. This number is checked when the connection is taken from the pool. When the connection pool reaper thread is enabled ('pg_pool_enable_reaper_thread' option), then this number is checked in background periodically.
 - `pg_pool_enable_reaper_thread` (default: `FALSE`): whether to enable the connection pool reaper thread, that periodically scans the pool to check the 'max_lifetime_millis' and 'idle_timeout_millis' and closes the connection which exceed the specified values. Either 'max_lifetime_millis' or 'idle_timeout_millis' must be set to a non-zero value for this option to be effective.
 - `pg_pool_health_check_query` (`VARCHAR`, default: `SELECT 1`): the query that is used to check that the connection is healthy. Setting this option to an empty string disables the health check.

Pool options that existed before, these options apply to both newly created pools and also to all existing pools:

 - `pg_connection_limit` (default: `max(num_cpus, 8)`): the same as `pg_pool_max_connections`
 - `pg_connection_cache` (default: `TRUE`): the same as `SET pg_pool_max_connections = 0` or `RESET pg_pool_max_connections`.

To change the configuration of an exiting pool new table function is added:

### Function: postgres_configure_pool

```sql
FROM postgres_configure_pool([<optional named parameters>]) -> TABLE
```

Allows to change the configuration of an existing connection pool (in an attached Postgres database). Returns the current state of a single pool or of all pools.

#### Parameters:

 - `catalog_name` (`VARCHAR`): the name (alias) of the attached Postgres database to which pool the configuration change is applied and details are returned. When `NULL` (default) returns the current state of pools for all attached catalogs without changing their configuration. Must be specified and non-NULL when any other option is specified.
 - `max_connections` (`UBIGINT`): maximum number of connections that are allowed to be cached in a connection pool for each attached Postgres database. This number can be temporary exceeded when parallel scans are used.
 - `wait_timeout_millis` (`UBIGINT`): maximum number of milliseconds to wait when acquiring a connection from a pool where all available connections are already taken.
 - `enable_thread_local_cache` (`BOOLEAN`): whether to enable the connection caching in thread-local cache. Such connections are getting pinned to the threads and are not made available to other threads, while still taking the place in the pool.
 - `max_lifetime_millis` (`UBIGINT`): maximum number of milliseconds the connection can be kept open. This number is checked when the connection is taken from the pool and returned to the pool. When the connection pool reaper thread is enabled ('enable_reaper_thread' argument), then this number is checked in background periodically.
 - `idle_timeout_millis` (`UBIGINT`): maximum number of milliseconds the connection can be kept idle in the pool. This number is checked when the connection is taken from the pool. When the connection pool reaper thread is enabled ('enable_reaper_thread' option), then this number is checked in background periodically.
 - `enable_reaper_thread` (`BOOLEAN`): whether to enable the connection pool reaper thread, that periodically scans the pool to check the 'max_lifetime_millis' and 'idle_timeout_millis' and closes the connection which exceed the specified values. Either 'max_lifetime_millis' or 'idle_timeout_millis' must be set to a non-zero value for this option to be effective.
 - `health_check_query` (`VARCHAR`): the query that is used to check that the connection is healthy. Setting this option to an empty string disables the health check.

#### Returns:

A table with 1 row for each pool with the following columns:

 - `catalog_name` (`VARCHAR`): the name (alias) of the attached Postgres database
 - `available_connections` (`UBIGINT`): the number of idle connection that are currently available in the pool
 - `max_connections` (`UBIGINT`): maximum number of connections that are allowed to be cached in the pool. This number can be temporary exceeded when parallel scans are used.
 - `wait_timeout_millis` (`UBIGINT`): maximum number of milliseconds to wait when acquiring a connection from a pool where all available connections are already taken
 - `thread_local_cache_enabled` (`BOOLEAN`): whether caching the connections caching in thread-local cache is enabled
 - `thread_local_cache_hits` (`UBIGINT`): the number of times connections were successfully acquired from a thread-local cache without going to the main pool
 - `thread_local_cache_misses` (`UBIGINT`): the number of times connections were not available in a thread-local cache and were taken from the main pool instead
 - `max_lifetime_millis` (`UBIGINT`): maximum number of milliseconds the connection can be kept open
 - `idle_timeout_millis` (`UBIGINT`): maximum number of milliseconds the connection can be kept idle in the pool
 - `reaper_thread_running` (`BOOLEAN`): whether the pool reaper thread is running; this thread periodically scans the pool to check the 'max_lifetime_millis' and 'idle_timeout_millis' and closes the connection which exceed the specified values
 - `health_check_query` (`VARCHAR`): the query that is used to check that the connection is healthy

#### Examples:

Change 2 options on a pool for the specified attached database:

```sql
ATTACH '...' AS my_db (TYPE POSTGRES);
FROM postgres_configure_pool(catalog_name='my_db', max_lifetime_millis=42000, enable_reaper_thread=TRUE);
```

Return current pool state for all attached databases:

```sql
FROM postgres_configure_pool();
```

Testing: new tests added for new configuration options and for `postgres_configure_pool` function.

### DIFF
--- a/src/include/storage/postgres_catalog.hpp
+++ b/src/include/storage/postgres_catalog.hpp
@@ -8,6 +8,8 @@
 
 #pragma once
 
+#include <memory>
+
 #include "duckdb/catalog/catalog.hpp"
 #include "duckdb/common/enums/access_mode.hpp"
 #include "postgres_connection.hpp"
@@ -79,6 +81,10 @@ public:
 
 	PostgresConnectionPool &GetConnectionPool() {
 		return *connection_pool;
+	}
+
+	shared_ptr<PostgresConnectionPool> GetConnectionPoolPtr() {
+		return connection_pool;
 	}
 
 	void ClearCache();
--- a/src/include/storage/postgres_connection_pool.hpp
+++ b/src/include/storage/postgres_connection_pool.hpp
@@ -52,4 +52,9 @@ private:
 	std::string health_check_query;
 };
 
+class PostgresConfigurePoolFunction : public TableFunction {
+public:
+	PostgresConfigurePoolFunction();
+};
+
 } // namespace duckdb
--- a/src/include/storage/postgres_connection_pool.hpp
+++ b/src/include/storage/postgres_connection_pool.hpp
@@ -26,7 +26,7 @@ using PostgresPoolConnection = dbconnector::pool::PooledConnection<PostgresConne
 
 class PostgresConnectionPool : public dbconnector::pool::ConnectionPool<PostgresConnection> {
 public:
-	PostgresConnectionPool(PostgresCatalog &postgres_catalog);
+	PostgresConnectionPool(PostgresCatalog &postgres_catalog, ClientContext &context);
 
 public:
 	bool TryGetConnection(PostgresPoolConnection &connection);
--- a/src/postgres_extension.cpp
+++ b/src/postgres_extension.cpp
@@ -69,6 +69,13 @@ static void SetPostgresConnectionLimit(ClientContext &context, SetScope scope, V
 	}
 	auto &config = DBConfig::GetConfig(context);
 	config.SetOption("pg_connection_limit", parameter);
+
+	// propagate the value also to 'pg_pool_max_connections'
+	optional_ptr<const ConfigurationOption> option;
+	auto setting_index = config.TryGetSettingIndex("pg_pool_max_connections", option);
+	if (setting_index.IsValid()) {
+		context.config.user_settings.SetUserSetting(setting_index.GetIndex(), parameter);
+	}
 }
 
 static void DisablePool(ClientContext &context, SetScope scope, Value &parameter) {
@@ -76,6 +83,8 @@ static void DisablePool(ClientContext &context, SetScope scope, Value &parameter
 		throw InvalidInputException("pg_connection_cache can only be set globally");
 	}
 	if (parameter.IsNull() || BooleanValue::Get(parameter)) {
+		Value def_size = Value::UBIGINT(PostgresConnectionPool::DefaultPoolSize());
+		SetPostgresConnectionLimit(context, scope, def_size);
 		return;
 	}
 	Value zero = Value::UBIGINT(0);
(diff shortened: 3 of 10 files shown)


basanthjenuhb:
@staticlibs 
Would you be able to update the documentation here as well for `postgres_configure_pool` function
https://duckdb.org/docs/current/core_extensions/postgres

staticlibs:
@basanthjenuhb 

Yes, the connection pool documentation is intended to be added there. Just the implementation is still being changed, and also the proper doc page requires quite a bit of writing. Besides the options reference it also should cover:

1. Why the pooling is needed - Potsgres' process per connection, DuckDB parallel scans
2. Pool tuning scenarios for PgBouncer in "transaction" mode and for direct Postgres
3. How the parallel scans being restricted by the pool size - additional query workers won't run when there are no free slots in the pool
4. Thread-local cache intended use (connection pinning  to DB instance worker pool threads) and various gotchas about it: (**NOT** cleaned up by the reaper thread, can starve the pool - should not be used with many client threads)

PR #432, opened 2026-04-07
Fix: populate PostgresType.oid for array columns to enable binary COPY

### LINKED ISSUE #431: Binary COPY is never used for tables with array columns (PostgresType.oid is never populated)
## Summary

When inserting into a PostgreSQL table that contains **any** array column (e.g. `varchar[]`, `text[]`, `int[]`), DuckDB's postgres extension always falls back to `FORMAT TEXT` for COPY, even when binary COPY should work. This results in ~2-2.5x slower bulk write throughput.

## Root Cause

In `postgres_table_entry.cpp`, `GetCopyFormat` calls `CopyRequiresText` for each column. For `LIST` types, the check is:

```cpp
case LogicalTypeId::LIST: {
    D_ASSERT(pg_type.children.size() == 1);
    auto &child_type = ListType::GetChildType(type);
    if (child_type.id() != LogicalTypeId::LIST) {
        if (!PostgresUtils::SupportedPostgresOid(child_type)) {
            return true;
        }
        if (pg_type.children[0].oid != PostgresUtils::ToPostgresOid(child_type)) {
            return true;  // <-- always triggers
        }
    }
    ...
}
```

The problem is that `PostgresType.oid` is **initialized to 0 and never populated** — neither `CreateEmptyPostgresType` (used for DDL-created tables) nor `TypeToLogicalType` (used for catalog-read tables) ever assigns the `oid` field on child types.

So the check always evaluates as:
```
0 != PostgresUtils::ToPostgresOid(child_type)  →  0 != 1043  →  true  →  FORMAT TEXT
```

This means **every table with any array column** silently falls back to TEXT format, regardless of whether binary would work correctly.

## Reproduction

```sql
-- Attach a PostgreSQL database
ATTACH 'postgresql://...' AS pg (TYPE postgres);

-- Create a table with a varchar[] column
CREATE TABLE pg.public.test_array (
    id INTEGER,
    name VARCHAR,
    tags VARCHAR[]
);

-- Insert data — this will use FORMAT TEXT internally
INSERT INTO pg.public.test_array
SELECT 1, 'test', ['a', 'b']::VARCHAR[];
```

Observe in PostgreSQL logs:
```
COPY "public"."test_array" FROM STDIN (FORMAT TEXT, NULL '')
```

Expected:
```
COPY "public"."test_array" FROM STDIN (FORMAT BINARY)
```

## Context

This was introduced in the fix for issue #206 (commit `4b227728`, April 2024 / v0.10.2), which added the OID comparison to prevent a crash when binary COPY wrote `VARCHAROID` (1043) into a `text[]` column (which expects `TEXTOID` = 25). The fix correctly identified the OID mismatch problem, but the comparison relies on `PostgresType.oid` which is never set.

## Suggested Fix

Populate `PostgresType.oid` from the PostgreSQL catalog when reading table metadata, and from `ToPostgresOid()` when creating tables via DDL. For the `text[]` vs `varchar[]` case specifically, the binary writer should use the target column's actual element OID rather than always emitting `VARCHAROID`.

## Impact

Bulk writes to PostgreSQL tables with array columns are ~2-2.5x slower than they should be. In our case, ETL throughput dropped from ~300 MB/s to ~130 MB/s when upgrading past v0.10.2.

## Environment

- DuckDB: v1.5.1
- postgres_scanner extension: v1.5.1 (commit c0e9256)
- PostgreSQL: Neon (PostgreSQL 17)
- OS: macOS (ARM64)

onhate:
## Summary

Fixes #431

`PostgresType.oid` was initialized to 0 and never populated for array child types, causing `CopyRequiresText` to always fall back to `FORMAT TEXT` for any table with array columns. This resulted in ~2x slower bulk write throughput.

## Root Cause

The check in `CopyRequiresText` for LIST types:
```cpp
if (pg_type.children[0].oid != PostgresUtils::ToPostgresOid(child_type)) {
    return true;  // always triggers because oid is always 0
}
```

Always evaluated as `0 != 1043 → true → FORMAT TEXT` because the `oid` field was never set in either code path.

### 1. `CreateEmptyPostgresType` (DDL-created tables)
When DuckDB creates a table via DDL (e.g. `CREATE TABLE ... VARCHAR[]`), the child `PostgresType.oid` was left at 0. Fixed by setting it via `ToPostgresOid()`, which is correct since DuckDB controls the DDL and the Postgres types will match.

### 2. `TypeToLogicalType` (catalog-read tables)  
When reading existing Postgres tables from the catalog, the child `PostgresType.oid` was also never set. Here `ToPostgresOid()` cannot be used because multiple Postgres type names map to the same DuckDB `LogicalType` (e.g. both `text` and `varchar` map to `VARCHAR`, but have different OIDs: 25 vs 1043). Added `TypeNameToPostgresOid()` to resolve the OID from the actual Postgres type name, preserving the text/varchar distinction.

### Result
- `varchar[]` columns now correctly use binary COPY in both paths
- `text[]` columns correctly fall back to TEXT copy (since `TEXTOID != VARCHAROID` and the binary writer always emits `VARCHAROID`)
- Other supported array types (`int[]`, `boolean[]`, `uuid[]`, etc.) also benefit from binary COPY

## Test plan
- [x] New test `attach_array_binary_copy.test` covering 4 scenarios:
  - DDL-created table with `VARCHAR[]` and `INT[]`
  - DDL-created table with `BOOLEAN[]`, `DATE[]`, `DOUBLE[]`, `UUID[]`
  - Existing `varchar[]` table created via `postgres_execute`
  - Existing `text[]` table created via `postgres_execute` (verifies TEXT fallback still works)
- [x] All existing array and binary copy tests pass

### DIFF
--- a/src/include/postgres_utils.hpp
+++ b/src/include/postgres_utils.hpp
@@ -67,6 +67,7 @@ public:
 	static string TypeToString(const LogicalType &input);
 	static string PostgresOidToName(uint32_t oid);
 	static uint32_t ToPostgresOid(const LogicalType &input);
+	static uint32_t TypeNameToPostgresOid(const string &type_name);
 	static bool SupportedPostgresOid(const LogicalType &input);
 	static LogicalType RemoveAlias(const LogicalType &type);
 	static PostgresType CreateEmptyPostgresType(const LogicalType &type);
--- a/src/postgres_utils.cpp
+++ b/src/postgres_utils.cpp
@@ -72,6 +72,45 @@ LogicalType PostgresUtils::RemoveAlias(const LogicalType &type) {
 	}
 }
 
+uint32_t PostgresUtils::TypeNameToPostgresOid(const string &type_name) {
+	if (type_name == "bool") {
+		return BOOLOID;
+	} else if (type_name == "int2") {
+		return INT2OID;
+	} else if (type_name == "int4") {
+		return INT4OID;
+	} else if (type_name == "int8") {
+		return INT8OID;
+	} else if (type_name == "float4") {
+		return FLOAT4OID;
+	} else if (type_name == "float8") {
+		return FLOAT8OID;
+	} else if (type_name == "varchar") {
+		return VARCHAROID;
+	} else if (type_name == "text") {
+		return TEXTOID;
+	} else if (type_name == "bytea") {
+		return BYTEAOID;
+	} else if (type_name == "date") {
+		return DATEOID;
+	} else if (type_name == "time") {
+		return TIMEOID;
+	} else if (type_name == "timestamp") {
+		return TIMESTAMPOID;
+	} else if (type_name == "timestamptz") {
+		return TIMESTAMPTZOID;
+	} else if (type_name == "interval") {
+		return INTERVALOID;
+	} else if (type_name == "timetz") {
+		return TIMETZOID;
+	} else if (type_name == "bit") {
+		return BITOID;
+	} else if (type_name == "uuid") {
+		return UUIDOID;
+	}
+	return 0;
+}
+
 LogicalType PostgresUtils::TypeToLogicalType(optional_ptr<PostgresTransaction> transaction,
                                              optional_ptr<PostgresSchemaEntry> schema,
                                              const PostgresTypeData &type_info, PostgresType &postgres_type) {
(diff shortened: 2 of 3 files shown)


onhate:
### Local test results (macOS ARM64, PostgreSQL 17, DuckDB v1.5.1)

| Test | Assertions | Result |
|------|-----------|--------|
| `attach_array_binary_copy.test` (new) | 63 | ✅ Pass |
| `attach_use_binary_copy.test` | 22 | ✅ Pass |
| `attach_text_array.test` | 6 | ✅ Pass |
| `attach_types.test` | 401 | ✅ Pass |
| `attach_types_multidimensional_array.test` | 66 | ✅ Pass |

No regressions in existing array or binary copy tests.

staticlibs:
Thanks!

onhate:
that's cool! thanks @staticlibs

PR #434, opened 2026-04-08
Don't rollback on reset if not necessary

leo-altertable:
After starting to use the new pooling mechanism we observed an increased amount of rollbacks and warning messages about issuing a rollback outside of a transaction.

My diagnostic is that this behavior is caused by the unconditional ROLLBACK in the `Reset` of the connection, which is called on its release.


staticlibs (approved):
Thanks for testing this early! That was an oversight, I will try to squeeze this fix into 1.5.2.

PR #435, opened 2026-04-11
Don't use ctid scan for negative relpages (like during scanning a partitioned table).

simi:
Scanning a partitioned table caused an infinite hang because PostgreSQL sets relpages=-1 for relkind='p' tables. Cast to idx_t (uint64) gives pages_approx = 2^64-1, so the scanner loops over ~18 quintillion pages.

~Now throws NotImplementedException at bind time with a message that includes the relation kind and table name.~ Now fallback to 0 estimate as scanner already does in few other cases. Keep original relpages value (`idx_t` needs to be "downgraded" to `int64_t` to be able to hold original -1 value) in `approx_num_pages` and later during preparing scan, check for negative `approx_num_pages` and fallback to `use_ctid_scan = false` for now.

---

Initial step to provide support for partitioned tables, since those are broken as reported at https://github.com/duckdb/duckdb-postgres/issues/220.

I have plan to open set of smaller PRs to improve the situation ending with full support of partitioned tables. This is initial foundation - accepting the broken state and fallback to slower but working scan.

Example EXPLAIN (SELECT hangs):

```sql
lake D EXPLAIN (FORMAT JSON) SELECT * FROM pg.public.sales_revenues;

┌─────────────────────────────┐
│┌───────────────────────────┐│
││       Physical Plan       ││
│└───────────────────────────┘│
└─────────────────────────────┘
[
    {
        "name": "POSTGRES_SCAN",
        "children": [],
        "extra_info": {
            "Table": "sales_revenues",
            "Projections": [
                "id",
                "account_id",
                "sale_id",
                "report_id",
                "report_group_id",
                "organization_id",
                "contract_id",
                "configuration_id",
                "product_id",
                "product_asset_id",
                "asset_id",
                "base",
                "share",
                "royalty",
                "currency",
                "chain",
                "statement_id",
                "unit_amount"
            ],
            "Estimated Cardinality": "18446744073709551568"
        }
    }
```

`pg_class` returns

```sql
reporting_development=# SELECT relkind, relpages FROM pg_class WHERE relname = 'sales_revenues';
 relkind | relpages 
---------+----------
 p       |       -1

```

### DIFF
--- a/src/include/storage/postgres_table_entry.hpp
+++ b/src/include/storage/postgres_table_entry.hpp
@@ -35,7 +35,7 @@ struct PostgresTableInfo {
 	unique_ptr<CreateTableInfo> create_info;
 	vector<PostgresType> postgres_types;
 	vector<string> postgres_names;
-	idx_t approx_num_pages = 0;
+	int64_t approx_num_pages = 0;
 };
 
 class PostgresTableEntry : public TableCatalogEntry {
@@ -64,7 +64,7 @@ public:
 	//! We would in this case remap them to "ID" and "id:1", while postgres_names store the original names
 	vector<string> postgres_names;
 	//! The approximate number of pages a table consumes in Postgres
-	idx_t approx_num_pages;
+	int64_t approx_num_pages;
 };
 
 } // namespace duckdb
--- a/src/postgres_utils.cpp
+++ b/src/postgres_utils.cpp
@@ -481,6 +481,21 @@ PostgresVersion PostgresUtils::ExtractPostgresVersion(const string &version_str)
 	return result;
 }
 
+string PostgresUtils::RelkindToString(const string &relkind) {
+	if (relkind == "r") {
+		return "table";
+	} else if (relkind == "v") {
+		return "view";
+	} else if (relkind == "m") {
+		return "materialized view";
+	} else if (relkind == "f") {
+		return "foreign table";
+	} else if (relkind == "p") {
+		return "partitioned table";
+	}
+	return "relation (relkind=" + relkind + ")";
+}
+
 string PostgresUtils::QuotePostgresIdentifier(const string &text) {
 	return KeywordHelper::WriteOptionallyQuoted(text, '"', false);
 }
--- a/src/postgres_utils.cpp
+++ b/src/postgres_utils.cpp
@@ -481,21 +481,6 @@ PostgresVersion PostgresUtils::ExtractPostgresVersion(const string &version_str)
 	return result;
 }
 
-string PostgresUtils::RelkindToString(const string &relkind) {
-	if (relkind == "r") {
-		return "table";
-	} else if (relkind == "v") {
-		return "view";
-	} else if (relkind == "m") {
-		return "materialized view";
-	} else if (relkind == "f") {
-		return "foreign table";
-	} else if (relkind == "p") {
-		return "partitioned table";
-	}
-	return "relation (relkind=" + relkind + ")";
-}
-
 string PostgresUtils::QuotePostgresIdentifier(const string &text) {
 	return KeywordHelper::WriteOptionallyQuoted(text, '"', false);
 }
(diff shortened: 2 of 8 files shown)


src/postgres_utils.cpp
```diff
@@ -530,6 +481,21 @@ PostgresVersion PostgresUtils::ExtractPostgresVersion(const string &version_str)
 	return result;
 }
 
+string PostgresUtils::RelkindToString(const string &relkind) {
```
simi: Even patch is targeted to partitioned tables, same issue would be now for any relkind. This mapping in here is to provide nice error message only for user with safe fallback. It handles all `relkinds` loaded at https://github.com/duckdb/duckdb-postgres/blob/c89234f0b1985f4ee0f52f16e742a1ab2d4ae4f0/src/storage/postgres_table_set.cpp#L32.

src/storage/postgres_table_set.cpp
```diff
@@ -128,9 +136,10 @@ void PostgresTableSet::CreateEntries(PostgresTransaction &transaction, PostgresR
 			if (info) {
 				tables.push_back(std::move(info));
 			}
-			auto approx_num_pages = result.IsNull(row, 2) ? 0 : result.GetInt64(row, 2);
 			info = make_uniq<PostgresTableInfo>(schema, table_name);
-			info->approx_num_pages = approx_num_pages;
+			auto relpages = result.IsNull(row, 2) ? 0 : result.GetInt64(row, 2);
+			auto relkind = result.IsNull(row, 12) ? string() : result.GetString(row, 12);
+			info->approx_num_pages = GetScanRelpages(relpages, relkind, table_name);
```
simi: `GetScanRelpages` is planned to be kept even once partitioned tables are fully supported as a guard for unexpected raw values.

src/storage/postgres_table_set.cpp
```diff
@@ -169,7 +178,9 @@ unique_ptr<PostgresTableInfo> PostgresTableSet::GetTableInfo(PostgresTransaction
 	for (idx_t row = 0; row < rows; row++) {
 		AddColumnOrConstraint(&transaction, &schema, *result, row, *table_info);
 	}
-	table_info->approx_num_pages = result->GetInt64(0, 2);
```
simi, line 172: This and following `GetInt64` missed the null check. Was it intended? I have added it in my change.

simi:
I'll check on CI failures.

staticlibs:
@simi 

Just FYI, some details on local test runs: https://github.com/duckdb/duckdb-postgres/issues/431#issuecomment-4195778007

src/storage/postgres_table_set.cpp
simi: Should I offboard this to small separate PR?

simi:
@staticlibs thanks for the hint, by mistake I have run only new tests before submitting PR. Clearly this should not break `ATTACH`. I have reworked this batch to just fallback to `use_ctid_scan = false` for negative `relpages` (as partitioned table provides). I have added new commit to see the difference over original idea and updated the PR name and description. I can squash if needed. If welcomed, I'll continue working on better support of partitioned tables in other PRs.

src/include/storage/postgres_table_entry.hpp
```diff
@@ -35,7 +35,7 @@ struct PostgresTableInfo {
 	unique_ptr<CreateTableInfo> create_info;
 	vector<PostgresType> postgres_types;
 	vector<string> postgres_names;
-	idx_t approx_num_pages = 0;
+	int64_t approx_num_pages = 0;
```
simi, line 38: Changed from `idx_t` (uint64) to `int64_t` to allow storing -1 (PostgreSQL's sentinel value for partitioned tables where `relpages` is unknown). My assumption is that's safe, since even the max is halved - at 8KB/page that's still ~74 ZB, which is safely beyond any real table 🙏.

staticlibs: Thanks for the details!

src/storage/postgres_table_set.cpp
staticlibs: I thin it is fine to have it as part of this PR.

staticlibs (approved):
Thanks! Looks good to me.

PR #441, opened 2026-04-17
Build libpq dependency using vcpkg

staticlibs:


simi:
Just wondering, why are ports needed?

staticlibs:
Hi!

> Just wondering, why are ports needed?

The overlay port was needed to apply the patch to the portfile, otherwise the build was failing like in this issue - microsoft/vcpkg/issues/39752 . Additionally Windows helper headers are added there that are used by some code in the scanner.

PR #453, opened 2026-04-29
Feat/yugabyte

wdroste:


.github/workflows/IntegrationTests.yml
```diff
@@ -38,6 +38,119 @@ jobs:
           clang-format --dump-config
           make format-check
 
+  linux-yugabyte:
+    name: YugabyteDB Tests
+    needs: format-check
```
staticlibs, line 43: It would be better to make this dependent on `linux-tests`.

.github/workflows/IntegrationTests.yml
```diff
@@ -262,7 +375,7 @@ jobs:
         threadsan: [0, 1]
 
     env:
-      CMAKE_BUILD_PARALLEL_LEVEL: 2
+      CMAKE_BUILD_PARALLEL_LEVEL: 1
```
staticlibs, line 378: I think this change is not needed, the CI passes with `2`.

.github/workflows/IntegrationTests.yml
```diff
@@ -367,11 +480,8 @@ jobs:
     - name: Dependencies
       shell: bash
       run: |
-        choco install \
-          ccache \
-          make \
-          ninja \
-          -y --force --no-progress
+        choco install ccache make -y --force --no-progress
+        choco install ninja -y --force --no-progress || pip install ninja
```
staticlibs, line 484: I think this change is not needed, if adding `pip` fallback - this should be done starting from the main duckdb/duckdb repo.

src/include/storage/postgres_connection_pool.hpp
```diff
@@ -47,6 +47,8 @@ class PostgresConnectionPool : public dbconnector::pool::ConnectionPool<Postgres
 	static std::string AcquireModeToString(PostgresPoolAcquireMode mode);
 	static void ValidatePoolAcquireMode(ClientContext &context, SetScope scope, Value &parameter);
 
+	std::unique_ptr<PostgresConnection> CreateConnectionToHost(const string &host, int32_t port);
```
staticlibs, line 50: M, this is kinda confusing. Are this direct-to-host connections pooled/cached in any way? If not, this logic should go outside of the connection pool.

src/storage/postgres_insert.cpp
```diff
@@ -121,8 +130,12 @@ SinkResultType PostgresInsert::Sink(ExecutionContext &context, DataChunk &chunk,
 	}
 	connection.CopyChunk(context.client, gstate.copy_state, chunk, gstate.varchar_chunk);
 	gstate.insert_count += chunk.size();
-	if (!keep_copy_alive) {
-		// if we are can't keep the copy alive we need to restart the copy during every sink
+	gstate.rows_in_batch += chunk.size();
+	if (gstate.yb_rows_per_transaction > 0 && gstate.rows_in_batch >= gstate.yb_rows_per_transaction) {
```
staticlibs, line 134: Can we explicitly check for `PostgresInstanceType::YUGABYTE` here for clarity? If the instance type check is not convenient here - lets add a comment that this branch is Yugabyte only.

test/sql/storage/attach_yugabyte_basic.test
```diff
@@ -0,0 +1,120 @@
+# name: test/sql/storage/attach_yugabyte_basic.test
+# description: Basic YugabyteDB attach, scan, and detection tests
+# group: [storage]
```
staticlibs, line 3: Please move all Yugabyte tests into a separate `yugabyte` directory.

staticlibs (changes requested):
Hi, thanks for the PR! I think it is fine to have additional logic for Yugabyte as long as it does not affect vanilla Postgres logic (AFAICS current version is properly guarded by the instance type checks).

Though the list of commits in this PR is a bit too long, would you mind squashing them into a single commit and removing `Co-Authored-By: Claude` to not violate the the [org-wide LLM policy](https://github.com/duckdb/duckdb/blob/main/CONTRIBUTING.md#generative-ai-policy)?

src/include/storage/postgres_connection_pool.hpp
wdroste: Yes you can directly connect to x number of the tservers in the cluster.

test/sql/storage/attach_yugabyte_basic.test
wdroste: I'll see how/if I can merge them back, I created it separating because there were so many different additional tests and runtime.

PR #447, opened 2026-04-20
#412  Add  OAuth bearer token injection via PQsetAuthDataHook

0814celsus:
Implements the PG18 PQsetAuthDataHook API to inject OAuth bearer tokens during OAUTHBEARER authentication. This allows connecting to OAuth-enabled PostgreSQL servers with a pre-obtained token, bypassing the device authorization flow.

Token sources (checked in priority order):
1. DuckDB setting: SET pg_oauth_token = '<token>'
2. Environment variable: PGOAUTHTOKEN

The hook is registered once at extension load time.
@staticlibs  let me know if anything else is needed here?

PR #456, opened 2026-04-30
Add OAuth bearer token injection via PQsetAuthDataHook

staticlibs:
This is a continuation of the PR #447, it makes the in-memory OAuth token session scoped, so multiple tokens can be used from the same DuckDB instance.


staticlibs:
@0814celsus

Hi! I wonder if you will be interested to take a look at this impl?

PR #461, opened 2026-05-05
Add postgres secret store

staticlibs:
This is a continuation of the PR #375.

The goal is to store the secrets in postgres so they can be retrieved remotely and automatically when attaching a lake. For example, the initial configuration is:

```sql
ATTACH 'postgres:postgresql://user:password@host/lake' AS metadata (SECRET_STORAGE_TABLE duckdb_secrets);
ATTACH 'ducklake:postgres:postgresql://user:password@host/lake' AS lake (DATA_PATH 's3://some-bucket');
CREATE SECRET IN postgres_metadata (
	TYPE S3,
	KEY_ID 'test_key_id',
	SECRET 'test_secret_value',
	REGION 'us-east-1'
);
```

After which, duckdb can be run in the JDBC driver for BI tools with a connection string like this, and it will be able to connect to postgres and S3 without further configuration:

```
jdbc:duckdb:ducklake:postgres:postgresql://user:password@host/lake
```

Feedback appreciated!


staticlibs:
@aclowes

Hi! I wonder if you would like to look at this PR, it bring the PR #375 to the latest `main` with minimal changes (alternatively these changes can be applied to the original PR, if convenient).

aclowes:
@staticlibs thanks for picking this up! I thought maybe storing secrets in postgres wasn't an approved idea. If you would either make a PR into my PR, or pull my commits and open your own PR on top of them, that would fix the attribution 😄

staticlibs:
@aclowes 

Thanks for the original PR impl!

> I thought maybe storing secrets in postgres wasn't an approved idea

There were concerns about storing the secrets in Postgres table unencrypted, but the idea itself were praised. Current expectation is that the access control is done by Postgres means, and that the encryption can be added later.

Updated the PR bringing in your original commits onto the `main` branch.

Added a change making the whole feature opt-in - requiring to specify the secrets table name in `SECRETS_TABLE` option to `ATTACH`. UX input is welcome!

I would think the `SECRETS_TABLE` needs to be added to the DuckLake extension with some hardcoded name, like `ducklake_secrets` to make the direct attach URLs working.

aclowes:
@staticlibs thanks for the rebase. Looks good. If you'd like an alternative to setting a default table name in DuckLake, perhaps in this PR you could provide a default table name, allow it to be overridden, and not make the table until someone inserts into a secret into it. I suspect there are objections to creating the table automatically, but since saving the secret to postgres requires opt-in already (`CREATE SECRET IN postgres_metadata`) perhaps lazy table creation is sufficient.

aclowes (approved):

staticlibs:
@aclowes 

Thanks for looking at this!

Opt-in/opt-out/lazy-init logic with the requirement to support DuckLake direct attach URLs (so no init phase at all) currently looks like a can of worms. I am updating the table creation to support read-only PG users now. Going to look at the lazy init after that. UX comments, especially for BI usage, are much appreciated!

staticlibs:
@aclowes 

> perhaps in this PR you could provide a default table name, allow it to be overridden

Just FYI, implemented this part. Though there is no lazy-init - the table (that may be default one) is probed in DB. It only created on init when `SECRET_STORAGE_TABLE duckdb_secrets` is specified explicitly on `ATTACH`. If table is not found/created - then the secret storage is not registered.

aclowes:
@staticlibs thanks for implementing this!

PR #475, opened 2026-05-25
Scan relation row types as STRUCT (not VARCHAR)

### LINKED ISSUE #474: VIEW with table-type columns returns VARCHARs in DuckDB
### What happens?

I have a view in PostgreSQL which returns columns of table's row type. For example:
```sql
-- PostgreSQL
CREATE TABLE public.users (name text, age int);
CREATE VIEW my_view as SELECT u.*::public.users from public.users as u;
```

After connecting to postgres from DuckDB and selecting from this view, I would expect a struct like this to be returned: `struct("name" varchar, age integer)`. Instead, a `varchar` type is returned.

For example:
```sql
-- DuckDB
install postgres;
load postgres;
attach ... as my_pg (type postgres);

select * from my_pg.public.my_view;
```
outputs to:
```
┌────────────────────────┐
│         users          │
│        varchar         │
├────────────────────────┤
│       (adam, 20)       │
└────────────────────────┘
```

This surprised me because selecting the whole row directly from the table like `select u from my_pg.public.users as u` outputs a struct type as expected. So I guess this must be a VIEW issue specifically.

### To Reproduce

As explained above. Create some table a view in PG:
```sql
-- PostgreSQL
CREATE TABLE public.users (name text, age int);
CREATE VIEW my_view as SELECT u.*::public.users from public.users as u;
```

Connect in DuckDB and try to select from the view:
```sql
-- DuckDB
install postgres;
load postgres;
attach ... as my_pg (type postgres);

select * from my_pg.public.my_view;
```

### OS:

MacOS 26.5 (25F71)

### PostgreSQL Version:

17.0

### DuckDB Version:

v1.5.3/v1.5.2/v1.5.0

### DuckDB Client:

CLI executable

### Full Name:

Adam Cholewiński

### Affiliation:

Dialo

- [x] I agree

- [x] I agree

wokalski:
Fixes #474 

This PR allows to use composite types from views/materialized views etc. Until now it was only possible for tables.

In postgres:
```sql
CREATE SCHEMA row_types;
CREATE TABLE row_types.widget(id INT, label VARCHAR);
CREATE VIEW row_types.widget_row AS SELECT w FROM row_types.widget AS w
```

In duckdb:
```sql
SELECT w.id, w.label FROM row_types.widget_row
----
1	hello
```

### DIFF
--- a/src/storage/postgres_type_set.cpp
+++ b/src/storage/postgres_type_set.cpp
@@ -90,8 +90,10 @@ JOIN pg_catalog.pg_namespace n ON n.oid = t.typnamespace
 JOIN pg_class ON pg_class.oid = t.typrelid
 JOIN pg_attribute ON attrelid=t.typrelid
 JOIN pg_type sub_type ON (pg_attribute.atttypid=sub_type.oid)
-WHERE pg_class.relkind = 'c'
+WHERE pg_class.relkind IN ('c', 'r', 'v', 'm', 'f', 'p')
 AND t.typtype='c'
+AND pg_attribute.attnum > 0
+AND NOT pg_attribute.attisdropped
 ${CONDITION}
 ORDER BY n.oid, t.oid, attrelid, attnum;
 )";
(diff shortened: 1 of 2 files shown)


staticlibs (approved):
Thanks for the PR! It looks good to me.

PR #476, opened 2026-05-26
Resolve cross-schema row types to STRUCT (not VARCHAR)

adamchol:
Follow-up to #474 / #475. After PR #475, row types of relations are discovered and registered under their own schema, so the in-schema case works. But when a relation in schema B has a column whose type is the row type of a relation in schema A, PostgresUtils::TypeToLogicalType still looked up the type entry against the relation's schema (B). It finds nothing there and falls back to VARCHAR.

### DIFF
--- a/src/include/postgres_utils.hpp
+++ b/src/include/postgres_utils.hpp
@@ -20,6 +20,7 @@ class PostgresTransaction;
 struct PostgresTypeData {
 	int64_t type_modifier = 0;
 	string type_name;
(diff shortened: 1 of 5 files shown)


staticlibs:
Hi, thanks for the PR! I think it looks good, just there is an outage with GH actions right now so CI tests are not running. I'll trigger the tests run when the actions are back on.

staticlibs (approved):
Thanks!

Issue #454, opened 2026-04-29
Insert on conflict with conflict target incorrectly calculates constraint


PadenZach:
### What happens?

After `ALTER TABLE ... DROP COLUMN`, `INSERT ... ON CONFLICT (cols) DO NOTHING` against a Postgres-backed table fails with:

> Binder Error: The specified columns as conflict target are not referenced by a UNIQUE/PRIMARY KEY constraint or index

The dropped column is not part of the conflict target, and the underlying Postgres unique constraint still exists and still references the same columns.

### 1. Start Postgres in Docker

```bash
docker run -d --name pg-bug \
  -e POSTGRES_PASSWORD=postgres \
  -e POSTGRES_DB=postgres \
  -p 5432:5432 \
  postgres:17

# wait for it to come up
until docker exec pg-bug pg_isready -U postgres >/dev/null 2>&1; do sleep 0.5; done
```

### 2. Create a table where attnums won't stay 1..N after a drop

The bug needs the dropped column to be *physically before* one of the conflict-target columns, so its attnum gap shifts the mapping. Easiest way: create the columns in the order `d, e, a, b, c` so attnums are `d=1, e=2, a=3, b=4, c=5`, then put the unique constraint on `(a, b, c)` (conkey `{3,4,5}`), then drop `e`.

```bash
docker exec -i pg-bug psql -U postgres <<'SQL'
CREATE TABLE tbl (
    d INT,
    e INT,
    a INT,
    b INT,
    c INT,
    UNIQUE (a, b, c)
);
SQL
```

### 3. Confirm ON CONFLICT works *before* the drop

```bash
duckdb <<'SQL'
INSTALL postgres_scanner;
LOAD postgres_scanner;
ATTACH 'host=localhost port=5432 user=postgres password=postgres dbname=postgres' AS pg (TYPE POSTGRES);
INSERT INTO pg.public.tbl (a, b, c, d, e) VALUES (1, 2, 3, 4, 5)
    ON CONFLICT (a, b, c) DO NOTHING;
SELECT * FROM pg.public.tbl;
SQL
```

Expected: insert succeeds, one row returned.

### 4. Drop column `e` and re-run the same INSERT (without `e`)

```bash
docker exec -i pg-bug psql -U postgres -c "ALTER TABLE tbl DROP COLUMN e;"

duckdb <<'SQL'
INSTALL postgres_scanner;
LOAD postgres_scanner;
[...]

PR #477, opened 2026-06-03
test(#454): add test for improper conflict resolution

PadenZach:
# WIP

These tests demonstrate how the current conflict building mechanism can fail if a postgres table has had columns dropped or reordered in certain ways. As described in #454

Will follow up with fix once I sit down and dig into the problem and possible fixes a bit more.


staticlibs:
Closing in favour of #523

PR #489, opened 2026-06-23
Bump duckdb and fix connect

carlopi:


staticlibs:
Looks good to me! Curiously, I was adding a similar change to `mysql_query` at this exact time.

carlopi:
I was trying:
```sql
ATTACH 'host=database-1-instance-1.XXX.rds.amazonaws.com port=5432 dbname=postgres user=postgres sslmode=require password=' || getenv('DB_TOKEN') AS aurora (TYPE postgres);
CONNECT aurora EXECUTE CREATE USER myuser WITH PASSWORD 'secret';
```
and got hit by "No fields returned by query"

staticlibs:
Yep, the same thing happens with `RemoteExecute` pushdowns. Unlike ODBC scanner, where `odbc_query` is used for DML too, in other scanners `*_query` and `*_execute` functions are historically separate. I think it makes perfect sense to make `*_query` functions to return `Success:BOOL` on DML and to make `*_execute` functions to be pure aliases to them.

carlopi:
This should now be ready

staticlibs (approved):
Please merge as convenient!

Support for DML also works in `mysql_query` now, just the impl/logic is slightly different - going to make the output and behaviour consistent between the scanners.

PR #493, opened 2026-07-03
first attempt to enable use_information_schema for ATTACH

khepin:
PR following the conversation started at https://github.com/duckdb/duckdb-postgres/issues/490

I have excluded from the PR handling enums, composite types and indexes that are unfortunately reported differently by different DB vendors.

I see a couple options here:
- leave ignored for now
- try to auto-detect the actual DB this is talking to and pick the right queries to get the right info this way
- change from a `use_info_schema` param to a `pg_compatibility_mode_for=(spanner|redshift|cockroach|...)`

Open to any of those options. Though auto-detection is probably the most brittle.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ cmake-build-debug
 .vscode
 .cache
 duckdb_unittest_tempdir
+.envrc
+docker-compose.yml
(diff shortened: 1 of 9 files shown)


staticlibs:
Hi, thanks for the PR! I will take a look.

For formatting please use the following:

```bash
pip install black==25.1.0 clang_format==11.0.1 cmake-format==0.6.13
make format-fix
```

staticlibs (approved):
Thanks! I will add formatting and minor changes on top and backport this to 1.5.

khepin:
Thank you!!

PR #508, opened 2026-07-08
fix: avoid crash by setting default index constraint type

### LINKED ISSUE #507: Catalog index entry bug
### What happens?

`duckdb_indexes()` scanning an attached Postgres catalog can crash with an internal assertion:

```
INTERNAL Error: Assertion triggered in file "duckdb/src/parser/parsed_data/create_index_info.cpp" on line 65: constraint_type == IndexConstraintType::UNIQUE || constraint_type == IndexConstraintType::NONE
```

`PostgresIndexSet::LoadEntries` constructs a `CreateIndexInfo` for every Postgres index then lazily loads into the catalog. But it doesn't assign `constraint_type`. Since the field has no default it ends up being whatever garbage memory that happens to be sitting at the address at the time, which then flows straight into the catalog entry. Anything that later calls `ToSQL()`/`ToString()` on it asserts if the value isn't `UNIQUE` or `NONE`. That then causes a crash in some scenario

### To Reproduce

Attach any Postgres database with at least one index (a plain user table's own index, or just the attached database's system catalog indexes, i.e. `pg_indexes`) and call `SELECT * FROM duckdb_indexes()`. Whether it crashes depends on the process's memory state at the time `LoadEntries` runs, so it doesn't reproduce on every single run.

Found while running `catalog_scan_concurrent_clear_stress.test_slow` (added for #502). The test's  `concurrentloop` calls `duckdb_indexes()` alongside `pg_clear_cache()` and `duckdb_tables()`, and the assertion fired reliably enough to abort most runs before completion. It's unrelated to the concurrency. No ThreadSanitizer warning precedes the crash and removing only the `duckdb_indexes()` statement from a scratch copy of the test lets the test deterministically pass everytime. Reduced further by forcing `constraint_type` to an invalid value at construction time by reproducing with a single Postgres index and no concurrency at all.

### OS:

arch linux

### PostgreSQL Version:

16.14

### DuckDB Version:

v1.5.4-8377-g2daa4fc9a4

### DuckDB Client:

cli

### Full Name:

brandon freeman

### Affiliation:

hallow

- [x] I agree

- [x] I agree

hello-world-bfree:
Resolves #507

## Problem

`duckdb_indexes()` scanning an attached Postgres catalog can crash with an internal assertion:

```text
INTERNAL Error: Assertion triggered in file "duckdb/src/parser/parsed_data/create_index_info.cpp" on line 65: constraint_type == IndexConstraintType::UNIQUE || constraint_type == IndexConstraintType::NONE
```

`PostgresIndexSet::LoadEntries` constructs a `CreateIndexInfo` for every Postgres index then lazily loads into the catalog. But it doesn't assign `constraint_type`. Since the field has no default it ends up being whatever garbage memory that happens to be sitting at the address at the time, which then flows straight into the catalog entry. Anything that later calls `ToSQL()`/`ToString()` on it asserts if the value isn't `UNIQUE` or `NONE`. That then causes a crash in some scenario

## Fix

Explicitly initializing `constraint_type` to `NONE` right where the entry is built, the same place every other field on `info` is already set

```cpp
CreateIndexInfo info;
info.constraint_type = IndexConstraintType::NONE;
info.SetQualifiedName(...);
```

### DIFF
--- a/src/storage/postgres_index_set.cpp
+++ b/src/storage/postgres_index_set.cpp
@@ -39,6 +39,7 @@ void PostgresIndexSet::LoadEntries(ClientContext &context, PostgresTransaction &
 		auto table_name = result.GetString(row, 1);
 		auto index_name = result.GetString(row, 2);
 		CreateIndexInfo info;
+		info.constraint_type = IndexConstraintType::NONE;
 		info.SetQualifiedName(
 		    QualifiedName(info.GetQualifiedName().Catalog(), schema.name, info.GetQualifiedName().Name()));
 		info.table = Identifier(table_name);
--- a/test/sql/storage/attach_duckdb_indexes_constraint_type.test
+++ b/test/sql/storage/attach_duckdb_indexes_constraint_type.test
@@ -1,0 +1,35 @@
+# name: test/sql/storage/attach_duckdb_indexes_constraint_type.test
+# description: duckdb_indexes() must not crash scanning Postgres indexes (constraint_type uninitialized read)
+# group: [storage]
+
+require postgres_scanner
+
+require-env POSTGRES_TEST_DATABASE_AVAILABLE
+
+statement ok
+ATTACH 'dbname=postgresscanner' AS s (TYPE POSTGRES)
+
+statement ok
+DROP TABLE IF EXISTS s.constraint_type_tbl
+
+statement ok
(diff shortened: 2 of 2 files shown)


staticlibs (approved):
Thanks for the PR!

PR #514, opened 2026-07-10
fix: detect Postgres table catalog staleness

### LINKED ISSUE #512: Catalog cache doesn't detect external Postgres DDL changes
### What happens?

The Postgres catalog cache, `PostgresCatalogSet`, loads table/schema metadata once and holds it for the lifetime of the process (or until `pg_clear_cache()` is called). There's no automatic way to notice when the underlying Postgres tables actually changed. If another connection, or even the same connection on a later statement, does some DDL there's nothing to tell the cache its in-memory view is misaligned. The next query against that table uses the stale column/type info until the cache is eventually cleared from a separate action.

This isn't often an issue in everyday `duckdb-postgres` but it's _the_ contract that DuckLake depends on. DuckLake opens its own internal connection to the same attached Postgres database, a second `ClientContext`, not the user's, and expects catalog changes made through one path to become visible to the other without any manual cache management in-between. Right now that's not guaranteed. A `ClientContext` that primed its cache before another connection's DDL landed keeps serving the old shape indefinitely.

The cache should detect drift on its own by checking whether Postgres's own bookkeeping (`pg_class.xmin`, row identity) has moved since the last load, and only then pay the cost of reloading. That'll be correct without needing every caller to know when to invalidate.

### To Reproduce

Two `ClientContext`s against the same attached Postgres schema. Not two `ATTACH`es of the same alias since each `ATTACH` gets its own catalog cache:

1. `con1`: `ATTACH 'dbname=x' AS s (TYPE POSTGRES); SELECT * FROM duckdb_tables();` — primes `con1`'s cache with the current table list.
2. `con2`: `ATTACH 'dbname=x' AS s2 (TYPE POSTGRES); CREATE TABLE s2.new_tbl(i INT);` — DDL from a separate connection.
3. `con1`: `SELECT * FROM duckdb_tables();` again. This run does _not_ show `new_tbl`. There will be no error, it'll just be silently missing. Until `pg_clear_cache()` is called explicitly that is.

An sharper version: 

2. (alt) Drop one table and create another with a different name in the same schema between steps 1 and 3 instead. So the Postgres *row count* is unchanged. Any staleness check based on cardinality alone misses as well.

Yet another separate, yet narrower case: 

With the same connection, inside one explicit (non-autocommit) transaction; i.e. 

```sql
ATTACH 'dbname=x' AS s (type POSTGRES);

CREATE OR REPLACE TABLE s.t (i INTEGER);
INSERT INTO s.t VALUES (42);

BEGIN;

ALTER TABLE t ADD COLUMN j INTEGER;

SELECT j from s.t;

ROLLBACK;

SELECT j FROM s.t;
```

Postgres correctly reverts `xmin` on rollback, which can collide with the pre-transaction signature and leave the cache serving mid-transaction, never-committed state after the rollback completes

### OS:

arch linux

### PostgreSQL Version:

16.14

### DuckDB Version:

v1.5.4-8377-g2daa4fc9a4

### DuckDB Client:

cli

### Full Name:

brandon freeman

### Affiliation:

hallow

- [x] I agree

### Have you tried the steps to reproduce? Do they include all relevant data and configuratio
n? Does the issue you report still appear there?

- [x] I agree

hello-world-bfree:
Resolves #512

## Problem

The Postgres catalog cache, `PostgresCatalogSet`, loads table/schema metadata once and holds it for the lifetime of the process (or until `pg_clear_cache()` is called). There's no automatic way to notice when the underlying Postgres tables actually changed. If another connection, or even the same connection on a later statement, does some DDL there's nothing to tell the cache its in-memory view is misaligned. The next query against that table uses the stale column/type info until the cache is eventually cleared from a separate action.

This isn't often an issue in everyday `duckdb-postgres` but it's _the_ contract that DuckLake depends on. DuckLake opens its own internal connection to the same attached Postgres database, a second `ClientContext`, not the user's, and expects catalog changes made through one path to become visible to the other without any manual cache management in-between. Right now that's not guaranteed. A `ClientContext` that primed its cache before another connection's DDL landed keeps serving the old shape indefinitely.

The cache should detect drift on its own by checking whether Postgres's own bookkeeping (`pg_class.xmin`, row identity) has moved since the last load, and only then pay the cost of reloading. That'll be correct without needing every caller to know when to invalidate.

## Fix

`PostgresTableSet` is the only nested set that can independently reload. `PostgresIndexSet` and `PostgresTypeSet` can not. They throw without a pre-supplied batch-load slice. So `PostgresTableSet` gets a staleness query: `(oid, relname, xmin)` per table in the schema, string-compared against the signature captured at last load. Any add, drop, rename changes that set. A plain row-count check doesn't.

```cpp
string PostgresTableSet::GetStalenessQuery() const {
        return "SELECT pg_class.oid, relname, pg_class.xmin FROM pg_class "
               "JOIN pg_namespace ON relnamespace = pg_namespace.oid "
               "WHERE relkind IN ('r','v','m','f','p') AND pg_namespace.nspname = $1 "
               "ORDER BY pg_class.oid";
}
```

The check runs once per `TryLoadEntries` call, inside the existing `load_lock`. It does not run per-entry, nor per-query. When there's a mismatch it clears and reloads exactly the way an explicit `pg_clear_cache()` already does.

A bug surfaced during implementation: 

A connection's own writes; i.e., `CREATE TABLE`, `DROP`, etc., were not updating the signature, so the *next* access on that same connection saw its own committed write as drift, wiped the entry it just created, and reloaded from Postgres. So it lost anything DuckDB's `AddColumn` path doesn't reconstruct, default values probably being the most notable. Fixed by refreshing the signature from `CreateTable`, `DropEntry`, and`ReloadEntry` directly, not from the generic bulk-load path.

That fix in turn exposed a second, narrower bug: 

A same-connection `ALTER TABLE` inside an explicit `BEGIN ... ROLLBACK`. Postgres reverts `xmin` on rollback (correct MVCC behavior) which can make a mid-transaction signature match the pre-transaction one once the rollback lands. The cache then keeps serving state that was never actually committed. Fixed by staging the signature instead of writing it directly whenever a transaction is open, and only promoting it to the real signature on `Commit()`. `Rollback()` just discards the pending value:

```cpp
void PostgresCatalogSet::RefreshStalenessSignature(PostgresTransaction &transaction) {
        auto signature = ComputeStalenessSignature(*transaction.Query(GetStalenessQuery()));
        if (transaction.GetTransactionState() == PostgresTransactionState::TRANSACTION_STARTED) {
                transaction.StageStalenessSignature(*this, std::move(signature));
                return;
        }
        staleness_signature = std::move(signature);
}
```

The pending value lives on `PostgresTransaction` as `pending_signatures`, mirroring the existing `referenced_entries` pattern, not on the catalog set itself since the cache is shared across every connection and a single pending slot would let one connection's commit or rollback clobber another's still-open transaction. `Commit()` promotes everything staged; `Rollback()` just drops it. Nothing in this path calls `ClearEntries()`, so it doesn't touch the reload machinery at all.

> [!NOTE]
> `PostgresIndexSet` and `PostgresTypeSet` don't get automatic staleness detection with this PR. They can't reload standalone at all today (confirmed: `LoadEntries` throws `InternalException` without a pre-supplied result slice; Same for both). Fixing that is a much bigger lift. Scope here is only `PostgresTableSet`

Verified against a live Postgres 16 instance, full `test/sql/storage/*` suite (108 tests, 3873 assertions) run twice, zero regressions. New coverage: external DDL visibility across two `ClientContext`s, an identity-swap case that rules out a pure cardinality signal, and a two-connection staged-signature case confirming one connection's pending (uncommitted) signature can't be promoted or clobbered by a concurrent, unrelated connection's commit.

### DIFF
--- a/.github/workflows/IntegrationTests.yml
+++ b/.github/workflows/IntegrationTests.yml
@@ -345,6 +345,7 @@ jobs:
         PGPASSWORD: postgres
         PGSSLMODE: require
         POSTGRES_TEST_DATABASE_AVAILABLE: 1
+        SANITIZER_BUILD: 1
         LOCAL_EXTENSION_REPO: 'build/relassert/repository'
       run: |
         source ./create-postgres-tables.sh
--- a/README.md
+++ b/README.md
@@ -52,6 +52,19 @@ D SELECT * FROM postgres_db.uuids;
 
 For more information on how to use the connector, refer to the [Postgres documentation on the website](https://duckdb.org/docs/extensions/postgres).
 
+### Connection Pooling
+
+By default, the extension opens a new connection whenever none are idle. Behind a transaction-mode pooler (e.g. PgBouncer `pool_mode=transaction`), this can exceed the pooler's own pool size under load and hang. If that applies to you, cap and wait instead:
+
+```sql
+SET pg_pool_acquire_mode = 'wait';
+SET pg_pool_max_connections = 16;        -- ~80-90% of your pooler's pool size
+
+ATTACH 'host=pgbouncer-host port=6432 dbname=mydb' AS postgres_db (TYPE postgres);
+```
+
+Note that `SET` statements are global and only apply pre-`ATTACH`. To change pooling on an already-attached database, use `postgres_configure_pool(catalog_name=...)` instead. Leave `force` mode if you connect directly to Postgres or through a session-mode pooler.
+
 ### AWS RDS IAM Authentication
 
 The extension supports AWS RDS IAM-based authentication, which allows you to connect to RDS/Aurora PostgreSQL instances using IAM database authentication instead of static passwords. This feature automatically generates temporary authentication tokens using the AWS SDK.
--- a/src/include/postgres_result.hpp
+++ b/src/include/postgres_result.hpp
@@ -48,6 +48,10 @@ public:
 		D_ASSERT(res);
 		return PQntuples(res);
 	}
+	idx_t ColumnCount() {
+		D_ASSERT(res);
+		return PQnfields(res);
+	}
 	idx_t AffectedRows() {
 		auto affected = PQcmdTuples(res);
 		if (!affected) {
--- a/src/include/postgres_utils.hpp
+++ b/src/include/postgres_utils.hpp
@@ -63,6 +63,8 @@ public:
 	static PGconn *PGConnect(const string &dsn, const string &attach_path);
 
 	static bool UseInformationSchemaIntrospection(optional_ptr<ClientContext> context);
+	static bool StalenessQueryEnabled(ClientContext &context);
+	static string GetCustomStalenessQuery(ClientContext &context);
 	static string DataTypeToTypeName(const string &data_type);
 	static LogicalType ToPostgresType(const LogicalType &input);
 	static LogicalType TypeToLogicalType(optional_ptr<PostgresTransaction> transaction,
--- a/src/include/storage/postgres_catalog_set.hpp
+++ b/src/include/storage/postgres_catalog_set.hpp
@@ -43,12 +43,17 @@ protected:
 		return false;
 	}
 	//! Empty (default) means staleness is never checked once loaded
-	virtual string GetStalenessQuery() const {
+	virtual string GetStalenessQuery(ClientContext &context) const {
 		return string();
 	}
 	void TryLoadEntries(ClientContext &context, PostgresTransaction &transaction);
-	void RefreshStalenessSignature(PostgresTransaction &transaction);
+	void RefreshStalenessSignature(PostgresTransaction &transaction, bool use_transaction_connection);
 
+private:
+	//! Runs LoadEntries + RefreshStalenessSignature, must be called with load_lock held
+	void LoadEntriesLocked(ClientContext &context, PostgresTransaction &transaction);
+
+protected:
 public:
 	void PromoteStalenessSignature(string signature);
 
--- a/src/include/storage/postgres_catalog_set.hpp
+++ b/src/include/storage/postgres_catalog_set.hpp
@@ -42,7 +42,15 @@ protected:
 	virtual bool HasInternalDependencies() const {
 		return false;
 	}
+	//! Empty (default) means staleness is never checked once loaded
+	virtual string GetStalenessQuery() const {
+		return string();
+	}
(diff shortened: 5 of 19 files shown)


staticlibs:
Hi, thanks for the PR! I wonder if we can place the `staleness_query` behind an option? I would think there is a good chance the current query may not work with Postgres-wire-compatible DBs that otherwise work with duckdb-postgres.

I think something like this may be better, 

 - option `pg_staleness_query`
 - default value: empty string, no staleness checks
 - special value: `PG_CLASS_DEFAULT` (or something similar) - will use the current query hardcoded in sources
 - custom query may be provided by a user

Perhaps it may be better to have 2 options - a flag and an actual value. 

I would like to get this into 1.5.5, and my concern is to not break existing scenarios with non-Postgres servers.

staticlibs:
As a follow-up to previous comment, perhaps having it "off" by default is a bad idea, maybe tying it with `pg_use_information_schema_introspection` will be better:

 - option `pg_staleness_query`
 - current query by default
 - if `pg_use_information_schema_introspection` is set - and the `pg_staleness_query` is set to empty string
 - if `pg_use_information_schema_introspection` is reset - `pg_staleness_query` is reset to default

hello-world-bfree:
> As a follow-up to previous comment, perhaps having it "off" by default is a bad idea, maybe tying it with `pg_use_information_schema_introspection` will be better:
> 
> * option `pg_staleness_query`
> * current query by default
> * if `pg_use_information_schema_introspection` is set - and the `pg_staleness_query` is set to empty string
> * if `pg_use_information_schema_introspection` is reset - `pg_staleness_query` is reset to default

Thanks @staticlibs! Makes a ton of sense. Probably should've looked to `pg_use_information_schema_introspection` from the start.

The plain boolean is compelling:

  - new option pg_staleness_query_enabled
  - default = !pg_use_information_schema_introspection
  - reuses the existing `ClearCacheOnSetting` callback that `pg_use_information_schema_introspection` already uses, so flipping it at runtime clears the cache the exact same way
  - `GetStalenessQuery` returns `""` when disabled; same as the base-class default) when disabled

wdyt?

staticlibs:
M, can we still make the query itself configurable? Assuming someone will be running DuckLake on Postgres-compatible DB and will want to avail of the staleness checks.

staticlibs (approved):
Thanks for the update!

PR #515, opened 2026-07-11
Apply the text protocol opt-out on the postgres_query (CONNECT) path

carlopi:
CONNECT on Redshift failed unless the user manually did SET pg_use_text_protocol=true, while ATTACH + USE worked.

DuckDB rewrites a CONNECT passthrough statement into postgres_query() (PostgresCatalog::RemoteExecute). Its bind, PGQueryBind, never called PostgresScanFunction::PrepareBind - which is where every server capability decision is made, including the Redshift opt-out from the binary COPY protocol. As a result the CONNECT path always wrapped the query in COPY (...) TO STDOUT (FORMAT "binary"), which Redshift does not support.

Instead of duplicating the Redshift check on each bind path, make the protocol a property of the attached database:
- add a use_text_protocol ATTACH option, accepting TRUE / FALSE / 'auto' (the default), stored on the PostgresCatalog
- PostgresCatalog::UseTextProtocol() is now the single place resolving the protocol for a database: an explicit option wins, 'auto' falls back to the pg_use_text_protocol setting or to Redshift auto-detection
- PGQueryBind calls PrepareBind, so the CONNECT path resolves it as well

Also resolve the protocol at the start of PrepareBind: it was set after the ctid scan and max_threads decisions that depend on it, which happened to be harmless only because Redshift reports major version 8.

### DIFF
--- a/src/include/postgres_version.hpp
+++ b/src/include/postgres_version.hpp
@@ -14,6 +14,10 @@ namespace duckdb {
 
 enum class PostgresInstanceType { UNKNOWN, POSTGRES, AURORA, REDSHIFT };
 
+//! How scans decide between the binary COPY protocol and the text protocol.
+//! AUTO derives it from the server (the binary COPY protocol is not supported by Redshift).
+enum class PostgresTextProtocolMode { AUTO, TEXT, BINARY };
+
 struct PostgresVersion {
 	PostgresVersion() {
 	}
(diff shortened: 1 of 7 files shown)


staticlibs (approved):
Looks good to me! Ideally we would like to disconnect `COPY` and `binary` logic for the protocol, to be able to use a binary protocol without `COPY` and with parameters support. But that is not implemented yet, and `COPY==binary` logic matches the current implementation. So I suggest to keep the flag/helpers as they are now in this PR and update them later.

PR #516, opened 2026-07-11
Bump duckdb and absorb patches

carlopi:


staticlibs:
The failure looks like a new one, it seems something has changed around `varchar_array`. I suggest to exclude it from the `foreach` list with `TODO`.

PR #523, opened 2026-07-20
fix: resolve ON CONFLICT target after DROP COLUMN

### LINKED ISSUE #454: Insert on conflict with conflict target incorrectly calculates constraint

CatherineThompson:
Fixes #454

Since Postgres column attnums are immutable, dropping a column introduces a gap between the attnum and the logical column index. This PR introduces a map of postgres column attnums to duckdb logical index to be used when building UniqueConstraints from the postgres catalog.

### DIFF
--- a/src/include/storage/postgres_table_entry.hpp
+++ b/src/include/storage/postgres_table_entry.hpp
@@ -13,6 +13,7 @@
 #include "duckdb/catalog/catalog.hpp"
 #include "duckdb/catalog/catalog_entry/schema_catalog_entry.hpp"
 #include "duckdb/parser/parsed_data/create_table_info.hpp"
+#include "duckdb/common/unordered_map.hpp"
 #include "postgres_utils.hpp"
 
 namespace duckdb {
@@ -39,6 +40,7 @@ struct PostgresTableInfo {
 	vector<PostgresType> postgres_types;
 	vector<string> postgres_names;
 	int64_t approx_num_pages = 0;
+	unordered_map<int64_t, idx_t> attnum_to_logical;
 };
 
 class PostgresTableEntry : public TableCatalogEntry {
--- a/src/storage/postgres_table_set.cpp
+++ b/src/storage/postgres_table_set.cpp
@@ -128,6 +128,10 @@ void PostgresTableSet::AddColumn(optional_ptr<PostgresTransaction> transaction,
 		column.SetDefaultValue(std::move(expressions[0]));
 	}
 	auto &create_info = *table_info.create_info;
+	if (!result.IsNull(row, 7)) {
+		auto attnum = result.GetInt64(row, 7);
+		table_info.attnum_to_logical[attnum] = create_info.columns.LogicalColumnCount();
+	}
 	if (is_not_null) {
 		create_info.constraints.push_back(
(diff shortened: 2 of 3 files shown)


staticlibs (approved):

PR #524, opened 2026-07-20
Allow synchronous snapshot on replicas and in RDS/Aurora

jgmills98:
Bringing forward 1/2 of the changes from #235 without removing Postgres 9.6 version checks. As stated by OP, synchronized snapshots are supported on replicas starting with PostgreSQL 10, and they're also supported in RDS/Aurora.


staticlibs (approved):

PR #527, opened 2026-07-22
Pin the first scan connection in multi-scan streaming queries

jgmills98:
Multiple scans of the same attached Postgres catalog in a single query hit the streaming path where the optimizer sets `can_use_main_thread = false`. On that path each scan reads over its own pool connections rather than the shared transaction connection.

In `PostgresGlobalState::TryOpenNewConnection`. The first connection a scan opens goes through the `ForceGetConnection()` branch, and that branch returned without ever calling `PostgresScanConnect`. Every later pool connection does call `PostgresScanConnect(..., snapshot, ...)`, so it runs `BEGIN` followed by `SET TRANSACTION SNAPSHOT` and reads at the shared snapshot. The first connection skipped both and scanned in autocommit, picking up a fresh snapshot on each `COPY`. So the first worker in a scan reads at a different point in time than the rest of the workers.


staticlibs (approved):
Thanks for the PR! This is a good catch, will add this to 1.5 too.

PR #528, opened 2026-07-22
Drop postgres 9.6 support

jgmills98:
Bringing over the 2nd half of #235 to remove support for Postgres 9.6.


staticlibs (approved):
Thanks for the PR! I think it makes sense to do this for 2.0.

PR #530, opened 2026-07-24
Reject out-of-range Postgres timestamps

marknefedov:
## Summary

`postgres_scan` silently corrupts out-of-range timestamps instead of erroring. A PostgreSQL
value that's valid but exceeds DuckDB's range wraps around via unchecked integer overflow:

    294276-12-31 23:59:59  →  290279-12-14 (BC) 15:58:09.448384

This PR makes both directions of the conversion check bounds explicitly instead of
overflowing:

- **Reader**: reject Postgres binary timestamps when the epoch-offset addition overflows
  or lands on a DuckDB infinity sentinel, instead of returning the wrapped value.
- **Writer**: reject DuckDB timestamps outside PostgreSQL's finite range
  (`datatype/timestamp.h` `MIN_TIMESTAMP`/`END_TIMESTAMP`) instead of underflowing.
  Found via the existing all-types round-trip test, which was masking the bug — DuckDB's
  minimum timestamp underflowed into a high AD value on write, then overflowed back to
  the original BC value on read, so the round trip "passed" on a corrupted value. That
  fixture now round-trips through `4713 BC`, a value both engines support.
- **Tests**: new PostgreSQL 18 regression reproducing the exact rows from the linked
  issue — supported controls, `timestamp`, `timestamptz`, and COPY to Parquet.

Reported in Snowflake-Labs/pg_lake#439


staticlibs (approved):

PR #531, opened 2026-07-24
Drain libpq results before throwing

marknefedov:
`PostgresConnection::ExecuteQueries` threw a `std::runtime_error` as soon as it saw a result with an error status, while still inside the `PQgetResult()` polling loop. This left the connection's result queue undrained — libpq expects every result to be pulled with `PQgetResult()` until it returns `NULL` before the connection is reused. Bailing out early via an exception left stale/pending results on the connection, so the next query issued on that (pooled/reused) connection could get confused by leftover state.

In `src/postgres_connection.cpp`, the error path no longer throws immediately. Instead, it records the first error message and `continue`s the loop so all remaining results are drained via `PQgetResult()`. Only after the loop exits (i.e. `PQgetResult()` returns `NULL`) is the preserved error thrown.


staticlibs (approved):

PR #534, opened 2026-07-24
Add a blob IN filter pushdown test (depends on database-connector#25)

marknefedov:
> **Blocked on https://github.com/duckdb/database-connector/pull/25 — CI will be red until that merges and the submodule pointer is bumped.**

## What this is

The SQL-level regression test for the blob `IN` pushdown bug. The fix itself is one line in the `database-connector` submodule (duckdb/database-connector#25); only the test belongs in this repo.

## The bug

`FilterPushdown::TransformExpression` serialises `COMPARE_IN` constants with `identifier_config`, whose blob literal prefix and suffix are empty, while `EncodeBlob` appends the closing quote unconditionally. With this repo's configuration (`'\x` prefix and `::BYTEA` suffix, set in `src/postgres_filter_pushdown.cpp:37`) a blob `IN` list is emitted as:

```sql
"blob_col" IN (6470666B67', 6F74686572')
```

Postgres rejects it with `trailing junk after numeric literal`, so `WHERE blob_col IN (...)` against an attached table fails outright. The `=` path was never affected — it already used `constant_config`.

## The test

`test/sql/storage/filter_pushdown_blob_in.test` runs blob `IN` / `NOT IN` filters against an attached Postgres table, including a six element list (the binder is more likely to keep a longer list as a `COMPARE_IN` rather than rewriting it to an `OR` chain) and blobs containing quote and backslash bytes. With the bug present these fail with the Postgres syntax error rather than returning wrong rows.

## To land

1. Merge duckdb/database-connector#25
2. Bump the `database-connector` submodule pointer on this branch
3. CI goes green, undraft

### DIFF
--- a/test/sql/storage/filter_pushdown_blob_in.test
+++ b/test/sql/storage/filter_pushdown_blob_in.test
@@ -1,0 +1,61 @@
+# name: test/sql/storage/filter_pushdown_blob_in.test
+# description: Test pushing down IN filters over blob columns
+# group: [storage]
+
+require postgres_scanner
+
+require-env POSTGRES_TEST_DATABASE_AVAILABLE
+
+statement ok
+ATTACH 'dbname=postgresscanner' AS s (TYPE POSTGRES)
+
+statement ok
+SET pg_experimental_filter_pushdown=true
+
+statement ok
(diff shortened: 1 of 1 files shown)


staticlibs (approved):
Thanks for adding the test!

PR #543, opened 2026-08-05
Fix idle transaction timeout error reporting

### LINKED ISSUE #378: idle-in-transaction timeout termination not reported to end user
### What happens?

If you encounter an idle-in-transaction termination in psql the error message sent to the end user shows that this was in fact the reason why their connection was closed. In `duckdb-postgres` the reported error message is much more ambiguous and does not mention `idle-in-transaction` at all.

I think the issue has to do with [this](https://github.com/duckdb/duckdb-postgres/blob/6dabd956065e59b24d615fe03bab84d677fca6df/src/postgres_connection.cpp#L76C56-L76C76) line. Possibly there is more to the error message than what is returned by  the `PQresultErrorMessage` function.

### To Reproduce

To recreate first set up a postgres instance on localhost with on port 5432

duckdb-postgres
```python
> import duckdb
> duckdb.install_extension('postgres')
> duckdb.load_extension('postgres')
> duckdb.sql(f"ATTACH 'user=postgres host=localhost' AS db (TYPE postgres, READ_ONLY);")

> duckdb.sql("SELECT * FROM postgres_query('db', 'begin; SELECT pg_sleep(10);')")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
_duckdb.Error: Failed to execute query "ROLLBACK": server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
```

Expected (psql)
```
> begin;
BEGIN

> -- wait a few second

> select 1;
FATAL:  terminating connection due to idle-in-transaction timeout
server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
The connection to the server was lost. Attempting reset: Succeeded.
```

### OS:

linux

### PostgreSQL Version:

17.6

### DuckDB Version:

1.4.1

### DuckDB Client:

python

### Full Name:

Andrew Jackson

### Affiliation:

Andrew Jackson

- [x] I agree

- [x] I agree

xjn2005:
Fixes #378. Preserve connection-level PostgreSQL FATAL errors when query cleanup rolls back a terminated connection. Adds an idle-in-transaction timeout regression test.

### DIFF
--- a/src/postgres_connection.cpp
+++ b/src/postgres_connection.cpp
@@ -101,8 +101,8 @@ unique_ptr<PostgresResult> PostgresConnection::TryQuery(optional_ptr<ClientConte
 
 	if (ResultHasError(result)) {
 		if (error_message) {
-			*error_message = StringUtil::Format("Failed to execute query \"" + query +
-			                                    "\": " + string(PQresultErrorMessage(result)));
+			*error_message =
+			    StringUtil::Format("Failed to execute query \"" + query + "\": " + string(PQerrorMessage(GetConn())));
 		}
 		PQclear(result);
 		return nullptr;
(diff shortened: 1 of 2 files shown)


staticlibs (approved):
Thanks for the update! Looks good to me.

PR #544, opened 2026-08-05
add connect_display option

Tmonster:
DuckDB-aws has been adding extra functionality to connect to RDS/Redshift in an easier way. The only bit that is missing is overriding the connect_display when connecting to these services without using an `ATTACH`. The connect display name defaults to the uri, which can be ugly for sometimes

This PR will allow extensions that call postgres-attach to also override the display name if postgres instances are immediately `connect`ed to instead of `attached` first.

This option gives extensions the ability to override the display name for CONNECT statements (and `attach` statements if they want) with something a bit more readable.

Before in duckdb-aws
```
memory D connect 'redshift-attach-connect-test' (type redshift);
postgres://'redshift-attach-connect-test.ccg... D ... D 
```

After
```
memory D connect 'redshift-attach-connect-test' (type redshift);
redshift: redshift-attach-connect-test D  
```


staticlibs (approved):
Thanks!

PR #555, opened 2026-08-21
Fix two libpq resource-handling bugs on COPY error paths

sfc-gh-mslot:
Two libpq resource-handling bugs on the COPY paths, found while looking into Snowflake-Labs/pg_lake#415.

`PostgresBinaryReader::FetchNextBuffer()` leaks the `PQgetCopyData` buffer when the message is shorter than the 2-byte tuple count it has to contain. `PQgetCopyData` allocates even when it returns a length we cannot use, and `buffer` was only assigned after the length check, so `FreeBuffer()` never saw it. Assigning it first hands it to the existing ownership, which frees it on the way out of the throw.

The three COPY error paths pass a possibly null `PGresult` to `PQresultErrorMessage()`. That does not crash - libpq returns an empty string - but it is how `Failed to copy data: ` with nothing after it gets produced, and a null result there means the connection failed, so the reason is on the connection. `ExecuteQueries()` and `PostgresQueryBind()` already do the `result ? PQresultErrorMessage(result) : PQerrorMessage(conn)` thing, so this just makes the COPY paths match.

The same report also covered a missing `PQfinish()` in `PGConnect()` and an undrained pipeline in `ExecuteQueries()`; both are already fixed on main, so they are not in this PR. For what it is worth, the `PQfinish()` one measured at about 19 kB per failed connection attempt in a long-lived process, so it was worth having.

No test - neither path is reachable from a well-behaved server.

### DIFF
--- a/src/postgres_binary_reader.cpp
+++ b/src/postgres_binary_reader.cpp
@@ -59,12 +59,16 @@ bool PostgresBinaryReader::FetchNextBuffer() {
 		return false;
 	}
 
+	// take ownership of the buffer before validating it, so that a short read
+	// does not leak it - PQgetCopyData allocates even when it returns a length
+	// we cannot use
+	buffer = new_buffer;
+
 	// len -2 is error
 	// we expect at least 2 bytes in each message for the tuple count
 	if (!new_buffer || len < sizeof(int16_t)) {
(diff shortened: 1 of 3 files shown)


staticlibs (approved):
Thanks for the PR! Looks good to me! I will add it to 1.5 branch shortly.

